PR #1, opened 2024-09-23
feat: [FW-25] create custom not found page

barroro:
- Create a custom not found page with an illustration and Link to go back to the home page
- Fix responsive Footer


DanielMoraDC (approved):
Looks great 🚀

PR #3, opened 2024-09-24
feat: [FW-26] add responsive navbar

barroro:
- Allow different navbar menus depending on the screen size


DanielMoraDC:
https://github.com/user-attachments/assets/ad0170c9-74ed-43d6-b2e3-a941d65e7a1e

I see that the button appears on the right but then the menu opens on the left. Is that intended?

DanielMoraDC:
I see the nav header for unlogged users is displayed on every page for a tiny amount of time. Can we prevent that from happening? Video:

https://github.com/user-attachments/assets/4331a916-09a9-47f4-939e-ea97fea5dbc6

DanielMoraDC:
I see that there is an error when we are in a page and we try to navigate to the same page. Example video is in mobile model, but desktop has the same problem:

https://github.com/user-attachments/assets/7585db35-5c53-43cd-856f-e98f70a750cf

DanielMoraDC (approved):
Looks good, great job!

PR #6, opened 2024-10-09
fix: [FW-35] build & a11y errors

barroro:
- Fix some problems when running to build the app
    - Move the components folder in the home folder to the components folder in the root
- Fix a11y errors
    - Color contrasts
    - Missing titles in elements


src/pages/api/auth/[...nextauth].ts
```diff
@@ -16,6 +16,7 @@ export const authOptions = {
       },
     }),
   ],
+  secret: process.env.NEXTAUTH_SECRET,
```
Comment (DanielMoraDC) [inline src/pages/api/auth/[...nextauth].ts:19]: What is this secret for?

Comment (barroro) [reply src/pages/api/auth/[...nextauth].ts:19]: It's the secret needed to encrypt the next auth JWT to work properly with the Spotify Auth. I need to update the readme file.

Comment (DanielMoraDC) [reply src/pages/api/auth/[...nextauth].ts:19]: Let's keep that in a separate PR, then :)

Makefile
```diff
@@ -14,6 +14,16 @@ run-app:
...
 
+.PHONY: run-build
+run-build:
+	@echo "Building the frontend app..."
+	@export $(shell cat .env | xargs) && npm run build
+
+.PHONY: run-start
+run-start:
+	@echo "Starting the production build locally..."
+	@export $(shell cat .env | xargs) && npm run start
```
DanielMoraDC, line 25: Probably a nitpick: can reuse this somehow?

I imagine something like (ChatGPT version 😅 ):

```make
.PHONY: run-build run-start set-env

# Helper to export environment variables
set-env:
	@export $(shell cat .env | xargs)

.PHONY: run-build
run-build: set-env
	@echo "Building the frontend app..."
	npm run build

.PHONY: run-start
run-start: set-env
	@echo "Starting the production build locally..."
	npm run start
```

barroro: Done! I have already reused some parts of the MakeFile 👌

DanielMoraDC (approved):

PR #11, opened 2024-12-07
feat: [FW-54] add validations in the generate page form

barroro:
- Add `react-hook-form` and `zod` packages for the validation.
- Validate the form and allow navigate to the next step if all required fields are filled.
- Add `Form` component from Shadcn package.
- Modify and add new test cases.
- **NOTE:** The copy URL button is now working yet


DanielMoraDC:
Should we close this one in favor of https://github.com/DanielMoraDC/festwrap-ui/pull/10?

DanielMoraDC:
Should we close this one in favor of https://github.com/DanielMoraDC/festwrap-ui/pull/10?

__tests__/generate.test.tsx
```diff
@@ -66,6 +66,78 @@ describe("GeneratePlaylistPage", () => {
...
+      within(firstStepContent).getByText(/steps.step1.title/i)
+    ).toBeInTheDocument()
+
+    // Select the existing playlist option
+    const useExistingPlaylistRadio = screen.getByRole("radio", {
+      name: /steps.step1.form.useExistingPlaylist.title/i,
+    })
+    await userEvent.click(useExistingPlaylistRadio)
+
+    // Click next
```
DanielMoraDC, line 127: These comments makes me think that we probably want the code being wrapped into a function with a more explanatory name. Specially for those code snippets that are repeated. For example, these 4 lines of code are repeated at least 3 times in this document.

__tests__/generate.test.tsx
```diff
@@ -66,6 +66,78 @@ describe("GeneratePlaylistPage", () => {
     ).toBeInTheDocument()
   })
 
+  it('should display the error message when the "Create new playlist" option is selected and the playlist name is not filled', async () => {
+    render(<GeneratePlaylistPage {...staticTranslations} />)
+
+    const firstNavigationStepButton = screen.getByRole("button", {
```
DanielMoraDC, line 72: I see these first checks are repeated in all tests, which is redundant. This not only makes each tests check multiple things, which is inconvenient, but would also make all tests fail in case these first navigation steps fail.

src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -6,81 +9,141 @@ import { Stepper, StepList, Step, StepContent } from "@components/ui/Stepper"
...
+    name: z.string().optional(),
+    playlistSelected: z.string().optional(),
+    isPrivate: z.boolean(),
+    bands: z.array(z.number().min(1)).nonempty("At least one band is required"),
+  })
+  .superRefine((data, ctx) => {
+    if (data.playlistType === "new" && !data.name) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "Name is required for NEW role",
```
DanielMoraDC, line 28: What do we mean by role, here? Should we also use lowercaps for the playlist type? We can quote the string, instead.

src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -6,81 +9,141 @@ import { Stepper, StepList, Step, StepContent } from "@components/ui/Stepper"
...
+        path: ["name"],
+      })
+    }
+
+    if (data.playlistType === "existing" && !data.playlistSelected) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message: "Playlist is required for EXISTING role",
```
DanielMoraDC, line 36: Same as in https://github.com/DanielMoraDC/festwrap-ui/pull/11/files#r1899344268

src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -6,81 +9,141 @@ import { Stepper, StepList, Step, StepContent } from "@components/ui/Stepper"
...
+    ])
+    if (isStepValid) setCurrentStep((prev) => prev + 1)
   }
 
-  const handleNext = () => {
+  const handleBack = () => setCurrentStep((prev) => prev - 1)
+
+  const onSubmit = (values: FormSchemaType) => {
     setCurrentStep((prev) => prev + 1)
+    console.log(values)
```
DanielMoraDC, line 75: Should we delete this?

__tests__/generate.test.tsx
barroro: Good point, I'm going to change it.

src/types/Playlist.ts
```diff
@@ -0,0 +1,19 @@
+export enum PlaylistType {
+  New = 'new',
+  Existing = 'existing',
+}
+
+export interface NewPlaylist {
```
DanielMoraDC, line 7: Shouldn't this be types, instead? Given they are data structured they feel more suited.

__tests__/generate.test.tsx
```diff
@@ -1,208 +1,223 @@
...
-describe("GeneratePlaylistPage", () => {
-  it("should render the form with the first step displayed", async () => {
-    render(<GeneratePlaylistPage {...staticTranslations} />)
-
-    const firstNavigationStepButton = screen.getByRole("button", {
+  vi.resetAllMocks();
+  cleanup();
+});
+
+const clickToNextButton = async () => {
```
DanielMoraDC, line 38: I have a question: any motivation for using inline arrow functions instead of using the `function` keyword? The latter feels more readable but it is also true that I am pretty new to Typescript.

DanielMoraDC (approved):
Left a small extra comment, feel free to self-resolve and merge after that.

I also left a question which should not be blocking.

src/types/Playlist.ts
barroro: I'm not sure 🤔 , because it's like the shape of an object right? Depending on the selection you need a different contract. But if you feel more comfortable with types in this case, I can change it 👌

DanielMoraDC: I imagined types would be more ideal for specifying data structures and interfaces for behaviors, but I see they can be used interchengably quite often. I would say we stick to something and be consistent. I am reading this thread [here](https://stackoverflow.com/a/52682220) and leaning towards using types for things like this, basically due to how you can define types as union and tuples of other types. What do you think?

__tests__/generate.test.tsx
barroro: No, for no particular reason. I have been working this way for the last 3 years, but there are some advantages depending on the application. For me there are a few advantages:
- Arrow functions can't be accessed before the function is defined. This problem forces me to put the function at the top of the file, which is more readable for me.
- It's an easy way to make functions. If you don't need to use `this` or other things that the `function` has, for me this way is more readable.
But depending on the case, it is more correct to use the `function` way.

src/types/Playlist.ts
barroro: It's fine for me! I'm going to change the interface for the types instead 👌

__tests__/generate.test.tsx
DanielMoraDC: I see this in the [Google Style guide](https://stackoverflow.com/a/52682220):

```
Prefer function declarations over arrow functions or function expressions when defining named functions.
Arrow functions may be used, for example, when an explicit type annotation is required.
```

I think we could stick to that. What do you think? I am happy to resolve and do this in another task, but I think aligning on some standard can be good to better standardize from now on.

PR #12, opened 2024-12-07
feat: [FW-55] copy to clipboard playlist URL when click on the copy URL button

barroro:
- Create a new hook to copy to the clipboard
- Refactor UserMenu to use the new hook instead of the Clipboard API directly
- Use the new hook in the third step to copy URL


src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,25 @@
+import { useState } from 'react';
+
+type UseCopyToClipboardReturn = {
+  copy: (_text: string) => void;
+  isCopied: boolean;
+};
+
+const DELAY_TIME = 1000;
```
DanielMoraDC, line 8: I would prevent hardcoding this. I would suggest we move it as a default argument in the function to make it easier to maintain. We could hardcode it there or even load delays for each usage using env variables (nice to have, not needed).

src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function useCopyToClipboard(): UseCopyToClipboardReturn {
```
DanielMoraDC, line 10: The names makes me understand that the function is doing the actual copy. What about using something like `getClipboard`? We could then return a `Clipboard` that has the `copy` method and the `isCopied` state.

src/hooks/useCopyToClipboard.ts
barroro: I'm going to change it to be more flexible, but the approach was to give more semantics in the hook and also I was thinking that it didn't make sense to have different delays when you are copying on the clipboard. But I agree with the change 👌

src/hooks/useCopyToClipboard.ts
barroro: Yes, this type of method is called a hook and usually starts with the prefix `use'. Also, the name `getClipboard` doesn't describe what the function does, because you're putting the data into the clipboard and you're not getting anything in the function.

DanielMoraDC (approved):

PR #16, opened 2025-01-08
chore: FW-62 create base code to connect with the backend

DanielMoraDC:
# Description

This PR introduces the base classes required to connect with the backend. The immediate next changes will implement the integration between the frontend and the backend.

Opened this as a draft due to the dependent PR not being merged. Once merged, I will move this to `ready`.

### DIFF
--- a/src/lib/artists.ts
+++ b/src/lib/artists.ts
@@ -1,0 +1,16 @@
+export class Artist {
+  private name: string;
+  private imageUri: string | undefined = undefined;
+  constructor(name: string, imageUri?: string) {
+    this.name = name;
+    this.imageUri = imageUri;
+  }
+
+  getName(): string {
+    return this.name;
+  }
+
+  getImageUri(): string | undefined {
+    return this.imageUri;
+  }
+}
--- a/src/lib/clients/backend.ts
+++ b/src/lib/clients/backend.ts
@@ -1,0 +1,51 @@
+import { Artist } from '../artists';
+import { AuthClient } from './auth';
+import { HttpClient, Method } from './http';
+
+export interface BackendClient {
+  searchArtists(_token: string, _name: string, _limit: number): Promise<Artist>;
+}
+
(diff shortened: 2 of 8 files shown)


DanielMoraDC:
@barroro PR ready for review 🎉

src/lib/clients/backend.ts
```diff
@@ -0,0 +1,51 @@
...
+    return this.httpClient
+      .send({
+        url: `${this.url}/artists/search`,
+        method: Method.Get,
+        params: { name, limit },
+        headers: { ...authHeader, ...backendAuthHeader },
+      })
+      .then((response) =>
+        response.data.map(
+          (artist: any) => new Artist(artist.name, artist.imageUri)
```
barroro, line 36: Does it make sense to return the instance? Shouldn't we return the DTO?

src/lib/artists.ts
```diff
@@ -0,0 +1,16 @@
+export class Artist {
```
barroro, line 1: We could add `toDTO` method or any way to get just the properties.

src/lib/clients/backend.ts
DanielMoraDC: I wanted to keep it simple, for now. The Artist entity only has 2 properties, so I thought adding a separate object would add complexity and provide no advantage, for now. What do you think?

barroro: Sorry, I may not have made myself clear. When I think in flow, I see that you have to call the methods inside the class to get the primitives in the component, right? Because you return the instance. But anyway, for now it's fine, we'll see what we need to do when the whole flow is defined.

barroro (approved):
Good job! 🔥

src/lib/artists.ts
DanielMoraDC: Resolving, as discussed in the other comment ✅

PR #17, opened 2025-01-19
feat: FW-64 Implement artists endpoint

DanielMoraDC:
# Description

Adds an endpoint to retrieve artists in the frontend. Implementing a wrapper on the backend endpoint so we do not have to expose the backend server.

# Test

If we start the app:

```shell
make run-app
```

Then we can open a terminal and run:

```shell
curl 'http://localhost:3000/api/artists/search?name=Iron&limit=4&token=<token>'
```

Using a proper token, this returns:

```json
{
   "artists":[
      {
         "name":"Iron & Wine",
         "imageUri":"https://i.scdn.co/image/ab6761610000f1786554f29133d7e27979e009d7"
      },
      {
         "name":"Iron Maiden",
         "imageUri":"https://i.scdn.co/image/ab6761610000f178f9978ad4808f6f2723124d19"
      },
      {
         "name":"Ironmouse",
         "imageUri":"https://i.scdn.co/image/ab6761610000f1785db1ad11adfddbcb6a57567a"
      },
      {
         "name":"The Ironix",
         "imageUri":"https://i.scdn.co/image/ab6761610000f178001e75c6cf35c402ae42a348"
      }
   ],
   "message":"Artists successfully retrieved"
}
```

### DIFF
--- a/src/pages/api/artists/search.ts
+++ b/src/pages/api/artists/search.ts
@@ -1,4 +1,5 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
+import { z } from 'zod';
 import { HttpClient, HttpBaseClient } from '@/lib/clients/http';
 import { Artist } from '@/lib/artists';
 import { BackendClient, HTTPBackendClient } from '@/lib/clients/backend';
@@ -19,6 +20,29 @@ type ResponseData = {
   artists?: ResponseArtist[];
 };
 
+function searchQuerySchema(defaultLimit: number, maxLimit: number) {
+  const limitMessage = `"limit" should be a number in interval [1, ${maxLimit}]`;
+  return z.object({
+    name: z.string({
+      message: '"name" should be provided as a string',
+    }),
+    token: z.string({
+      message: '"token" should be provided as a string',
+    }),
+    limit: z
+      .number({
+        message: limitMessage,
+      })
+      .gt(1, {
+        message: limitMessage,
+      })
+      .lte(maxLimit, {
+        message: limitMessage,
+      })
+      .default(defaultLimit),
+  });
+}
+
 export function createSearchArtistHandler({
   client,
   defaultLimit,
@@ -28,40 +52,21 @@ export function createSearchArtistHandler({
     request: NextApiRequest,
     response: NextApiResponse<ResponseData>
   ): Promise<void> {
-    const { name, token, limit } = request.query;
-
-    let parsedLimit: number;
-    try {
-      parsedLimit = parseParamNumberWithDefault(
-        limit,
-        defaultLimit,
-        maxLimit,
-        'limit'
-      );
-    } catch (e) {
-      response.status(400).json({ message: (e as Error).message });
-      return;
-    }
-
-    if (typeof name !== 'string') {
+    const parsedArgs = searchQuerySchema(defaultLimit, maxLimit).safeParse(
+      request.query
+    );
+    if (!parsedArgs.success) {
       response
         .status(400)
-        .json({ message: '"name" should be provided as a string' });
-      return;
-    }
-
-    if (typeof token !== 'string') {
-      response
-        .status(400)
-        .json({ message: '"token" should be provided as a string' });
+        .json({ message: parsedArgs.error.errors[0].message });
       return;
     }
 
     try {
       const searchResults = await client.searchArtists(
-        token as string,
-        name as string,
-        parsedLimit
+        parsedArgs.data.token as string,
+        parsedArgs.data.name as string,
+        parsedArgs.data.limit
       );
       response.status(200).json({
         artists: searchResults.map((artist: Artist) => ({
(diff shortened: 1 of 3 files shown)


DanielMoraDC:
@barroro This PR is now ready to be reviewed 🎉

src/pages/api/artists/search.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+const defaultLimit: number = parseInt(
+  process.env.SEARCH_ARTISTS_DEFAULT_LIMIT || '5'
+);
+const maxLimit: number = parseInt(process.env.SEARCH_ARTISTS_MAX_LIMIT || '10');
+const serverHost: string = process.env.SERVER_HOST || 'http://localhost';
+const serverPort: number = parseInt(process.env.SERVER_PORT || '8080');
+const httpClient: HttpClient = new HttpBaseClient();
+const client: BackendClient = new HTTPBackendClient(
```
barroro, line 106: Should we move the instances to a place for reuse this code?

src/pages/api/artists/search.ts
```diff
@@ -0,0 +1,114 @@
...
+  name: string;
+  imageUri: string | undefined;
+};
+
+type ResponseData = {
+  message: string;
+  artists?: ResponseArtist[];
+};
+
+export function createSearchArtistHandler({
```
barroro, line 22: I have a question. I would like to understand what is a handler for you, because for me this part should be a controller. It's a bit confusing for me because when I use the `handler` name, it's for the event handlers when you're using microservices and you need to listen to each event.

src/pages/api/artists/search.ts
```diff
@@ -0,0 +1,114 @@
...
+        defaultLimit,
+        maxLimit,
+        'limit'
+      );
+    } catch (e) {
+      response.status(400).json({ message: (e as Error).message });
+      return;
+    }
+
+    if (typeof name !== 'string') {
```
barroro, line 46: We could use the `zod` package for all these validations. Because we already use it for the form validations and it allows you to have it more encapsulated. What do you think?
Check docu: [link](https://zod.dev/?id=basic-usage)

src/pages/api/artists/search.ts
DanielMoraDC: I kept it simple because it is the first API endpoint and did not want to add more complexity for this PR. I am happy to either do it here or in the next PR, as you prefer.

src/pages/api/artists/search.ts
DanielMoraDC: To me everything that reacts to an event (in this case an HTTP request) is a handler. I usually associate Controller with MVC architectures, which I do not think we are following. Either way works for me, I do not have a strong opinion.

src/pages/api/artists/search.ts
DanielMoraDC: I'll take a look at it!

DanielMoraDC: I checked the `zod` package. This is how the modified code looks like.

## Name and token validation

From this:

```ts
    if (typeof name !== 'string') {
      response
        .status(400)
        .json({ message: '"name" should be provided as a string' });
      return;
    }

    if (typeof token !== 'string') {
      response
        .status(400)
        .json({ message: '"token" should be provided as a string' });
      return;
    }
```

To this:

```ts
    if (!z.string().safeParse(name).success) {
      response
        .status(400)
        .json({ message: '"name" should be provided as a string' });
      return;
    }

    if (!z.string().safeParse(token).success) {
      response
        .status(400)
        .json({ message: '"token" should be provided as a string' });
      return;
    }
```

## Limit validation

From this:

```ts
    if (!z.number().gt(1).lte(maxValue).safeParse(parsed).success) {
      throw Error(
        `"${paramName}" should be a number in interval [1, ${maxValue}]`
      );
    }
```

To this:

```ts
    if (isNaN(parsed) || parsed < 1 || parsed > maxValue) {
      throw Error(
        `"${paramName}" should be a number in interval [1, ${maxValue}]`
      );
    }
```

I believe that it does not provide any benefit in terms of code simplicity or readability (actually, a bit less readable?). Given there is not clear benefit, I would rather keep the native code than creating extra coupling with the package.

What do you think?

barroro: No, you should use using the object method like this:
```
const schemaValidation = z.object({
  name: z.string('"name" should be provided as a string'),
  token: z.string('"token" should be provided as a string'),
  limit: z.number('"limit" should be provided as a string'),
});

try {
  // If not parse, it throws an error
  schemaValidation.parse(request.query)
} catch (error) {
  response
        .status(400)
        .json(error.message);
}
```

I'm not sure 100% if it works because I didn't try but is quite similar the approach.

DanielMoraDC: Done ✅

barroro (approved):

PR #18, opened 2025-01-19
feat(auth): refresh token when expired

DanielMoraDC:
# Description

Fixes auth so a refresh token is requested whenever login has expired. By doing so, we prevent using expired tokens in future calls.

Also renamed session attributes for consistency.

# References

- [Spotify docs](https://developer.spotify.com/documentation/web-api/tutorials/refreshing-tokens)
- [Authjs docs](https://authjs.dev/guides/refresh-token-rotation)

### DIFF
--- a/src/types/next-auth.d.ts
+++ b/src/types/next-auth.d.ts
@@ -4,7 +4,7 @@ declare module 'next-auth' {
   // eslint-disable-next-line no-unused-vars
   interface Session {
     user: {
-      access_token: string;
+      accessToken: string;
     } & DefaultSession['user'];
   }
 }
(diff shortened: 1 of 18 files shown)


src/types/next-auth.d.ts
```diff
@@ -4,7 +4,7 @@ declare module 'next-auth' {
   // eslint-disable-next-line no-unused-vars
   interface Session {
     user: {
```
barroro, line 6: We have to decide how we want to define the variables with camelCase or snakeCase, because I'm usually working with camelCase to have all files with the same structure. What do you think?

DanielMoraDC: We had the problem before that we were using the returned `token` as base object, which already contained some variables in snake case. I made it consistent, but it is true that now this is not consistent with the rest. Let me change it.

DanielMoraDC: Done ✅

barroro (approved):
Good job!

PR #19, opened 2025-01-29
chore: add axios types

barroro:
- Add the package to work axios as typed


DanielMoraDC:
I read from the [npm repository](https://www.npmjs.com/package/@types/axios) that axios already contains the definitions. Do we need this for newer versions of axios?

PR #21, opened 2025-02-04
chore: minor changes

DanielMoraDC:
# Description

Set of small changes. I moved them into a separate PR to ease review or incoming PRs.

### DIFF
--- a/src/components/generate/PlaylistSearchBandsForm/SearchBandsCombobox.tsx
+++ b/src/components/generate/PlaylistSearchBandsForm/SearchBandsCombobox.tsx
@@ -9,14 +9,14 @@ import {
 } from 'lucide-react';
 import Image, { StaticImageData } from 'next/image';
 
-interface Item {
+interface SearchedArtist {
   id: number;
   title: string;
   icon: StaticImageData;
 }
 
 interface SearchComboboxProps {
-  options: Item[];
+  options: SearchedArtist[];
   values: number[];
   onChange: (_values: number[]) => void;
   placeholder?: string;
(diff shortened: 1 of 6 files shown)


src/components/generate/PlaylistSearchBandsForm/SearchBandsCombobox.tsx
```diff
@@ -74,7 +63,18 @@ export function SearchBandsCombobox({
     setActiveIndex(-1);
   };
 
-  const handleItemSelect = (item: Item) => {
+  const handleOutsideClick = (event: MouseEvent) => {
```
barroro, line 66: Good approach! But it should be declared before where it's used because is an arrow function. I don't know why it works, but it shouldn't.

barroro (commented):
Good job! 🔥

src/components/generate/PlaylistSearchBandsForm/SearchBandsCombobox.tsx
DanielMoraDC: I believe it does not matter much, because the function is called on `useEffect` trigger, which I believe happens after the function has been called.

Anyway, I have moved all arrow functions to the top and kept all useEffects at the bottom so it is consistent. Let me now what you think.

barroro: I think the same as you, but checking the expected behaviour it should work this way. Good job 👌

barroro (approved):

PR #22, opened 2025-02-04
refactor: move search result to separate file

DanielMoraDC:
# Description

Refactoring search result list to make it more modular and easier to change/read.

### DIFF
--- a/src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
+++ b/src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
@@ -1,0 +1,50 @@
+import { CircleCheck } from 'lucide-react';
+import Image, { StaticImageData } from 'next/image';
+
+interface BandSearchResultProps {
+  key: number;
+  name: string;
+  isActive: boolean;
+  isSelected: boolean;
+  handleItemSelect: (_item: any) => void;
+  icon?: StaticImageData;
+}
+
+const BandSearchResult = ({
+  key,
+  name,
+  isActive,
+  isSelected,
+  handleItemSelect,
+  icon,
+}: BandSearchResultProps) => {
+  return (
+    <li
+      key={key}
+      role="option"
+      aria-selected={isActive}
+      className={`flex items-center px-4 py-2 cursor-pointer ${isActive ? 'bg-blue-100' : 'hover:bg-gray-100'}`}
+      onClick={handleItemSelect}
+    >
+      {icon ? (
+        <Image
+          src={icon.src}
+          alt=""
+          width={icon.height}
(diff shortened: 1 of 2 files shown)


src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -0,0 +1,50 @@
...
+  isSelected,
+  handleItemSelect,
+  icon,
+}: BandSearchResultProps) => {
+  return (
+    <li
+      key={key}
+      role="option"
+      aria-selected={isActive}
+      className={`flex items-center px-4 py-2 cursor-pointer ${isActive ? 'bg-blue-100' : 'hover:bg-gray-100'}`}
```
barroro, line 26: You could use the `cn` method for the conditional styles. There are some examples in the project.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -0,0 +1,50 @@
...
+  key: number;
+  name: string;
+  isActive: boolean;
+  isSelected: boolean;
+  handleItemSelect: (_item: any) => void;
+  icon?: StaticImageData;
+}
+
+const BandSearchResult = ({
+  key,
```
barroro, line 14: You don't need to define `key` prop, React automatically adds this prop for you.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -0,0 +1,50 @@
...
+  name,
+  isActive,
+  isSelected,
+  handleItemSelect,
+  icon,
+}: BandSearchResultProps) => {
+  return (
+    <li
+      key={key}
```
barroro, line 23: You can also remove this prop here. The `key` is used to render elements in a list and to identify each element. You only use this prop where the map is used (iteration).

barroro (commented):
I added some comments but good job creating new components! 🔥

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
DanielMoraDC: Good catch! I copied the arguments outside and did not pay enough attention. Fixed, now.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
DanielMoraDC: Fixed ✅

barroro (approved):

PR #23, opened 2025-02-06
refactor: improve http client backend flow

barroro:
- Create http auth builder class to build both headers (GCP, Spotify)
- Separate `HTTPBackendClient` by entities (Artists, Playlists etc...)
- Get access token using `getToken` from NextAuth


src/lib/clients/artists-backend.ts
```diff
@@ -1,24 +1,27 @@
 import { Artist } from '../artists';
 import { AuthClient } from './auth';
+import { BaseHTTPClientWithAuth } from './base-backend';
 import { HttpClient, Method } from './http';
 
-export interface BackendClient {
+export interface IArtistsHTTPBackendClient {
```
DanielMoraDC, line 6: The interface is not actually tied to HTTP, it could be something else (e.g. GRPC). I would name it like "ArtistsClient" or similar, instead.

src/lib/clients/artists-backend.ts
```diff
@@ -1,24 +1,27 @@
 import { Artist } from '../artists';
```
DanielMoraDC, line 1: I am thinking now that we could name this file simply `src/lib/clients/artists.ts`. The backend part of it does not provide any additional information and it just makes the name longer.

src/lib/clients/base-backend.ts
```diff
@@ -0,0 +1,24 @@
+import { AuthClient } from './auth';
```
DanielMoraDC, line 1: The name of the class/interface feels misleading. We reference HTTPClient but it actually just creates a header out of a given client. We can probably use a name such as `AuthHeaderBuilder` or similar. It feels more aligned with what it does.

Also, note that in the old backend there is [another header](https://github.com/DanielMoraDC/festwrap-ui/blob/main/src/lib/clients/backend.ts#L30) being used. The header `Authorizarion` header is expected to include the Spotify token and the other header includes the token for service to service authentication (required in GCP). It is optional because it is only needed within Google Cloud, but we need to bypass it for local testing. I think we can include both headers here to improve reusability of client code.


I think I would rather pass th

src/lib/clients/base-backend.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  beforeEach(() => {
+    authClient = {
+      getToken: vi.fn().mockResolvedValue('test-token'),
+      getHeaderName: vi.fn().mockReturnValue('Authorization'),
+    } as unknown as AuthClient;
+  });
+
+  it('should return an empty object if authClient is undefined', async () => {
+    const client = new BaseHTTPClientWithAuth();
```
DanielMoraDC, line 16: I like to separate the 3 main sections of my tests with a line break:
- Given: the test preparation of the context.
- When: the actual action.
- Then: result or expectations.

It helps me a lot to read the tests but also write them. I know these tests are short anyway, but I just wanted to find alignment for future PRs 😃  What do you think?

src/lib/clients/base-backend.test.ts
```diff
@@ -0,0 +1,33 @@
+import { BaseHTTPClientWithAuth } from './base-backend';
+import { AuthClient } from './auth';
+import { describe, beforeEach, it, expect, vi } from 'vitest';
+
+describe('BaseHTTPClientWithAuth', () => {
+  let authClient: AuthClient;
+
+  beforeEach(() => {
+    authClient = {
+      getToken: vi.fn().mockResolvedValue('test-token'),
```
DanielMoraDC, line 10: I usually resort to mocks only when implementing the fake becomes too cumbersome. I feel mocks are quite convoluted but more importantly they create extra coupling to the specific framework. Then you can use vitest to spy on the fake, but at least you reduce the usage of the mocking. What is your opinion on this?

src/lib/clients/artists-backend.ts
barroro: Perfect, it makes sense. Done!

src/lib/clients/artists-backend.ts
barroro: Done!

src/lib/clients/base-backend.test.ts
barroro: I have been working with given/when/then in the tasks but I've never used it in the tests. So, how do you normally work with that? Do you have an example to have a look at?

src/lib/clients/artists.ts
```diff
@@ -1,60 +1,51 @@
...
+export interface ArtistsClient {
   searchArtists(
     _token: string,
     _name: string,
     _limit: number
   ): Promise<Artist[]>;
 }
 
-export class HTTPBackendClient implements BackendClient {
+export class ArtistsHTTPBackendClient
```
DanielMoraDC, line 14: I would probably remove the 'Backend' part from the name, as it does not provide much information and makes the name longer.

src/lib/clients/base-backend.test.ts
DanielMoraDC: In practise, what I usually do is simply separate the lines that belong to each section with a linespace (example [here](https://github.com/DanielMoraDC/festwrap-ui/blob/main/src/lib/clients/backend.test.ts#L30)).

src/lib/clients/auth.ts
```diff
@@ -1,6 +1,6 @@
 import { HttpClient, Method } from './http';
 
-export interface AuthClient {
+export interface GCPAuthClient {
```
DanielMoraDC, line 3: I believe the AuthClient should be generic, so we can switch the implementation (e.g. GCP, AWS, Azure) seamessly.

src/lib/clients/auth.ts
```diff
@@ -36,7 +36,7 @@ export class GCPHTTPAuthClient {
   }
 }
 
-export class FakeAuthClient {
+export class FakeGCPAuthClient {
```
DanielMoraDC, line 39: Same here, I think this should be `FakeAuthClient` without the GCP.

src/lib/clients/auth.ts
```diff
@@ -65,3 +65,51 @@ export class FakeAuthClient {
     return this.header;
   }
 }
+
+export interface AuthHeaderBuilder {
```
DanielMoraDC, line 69: When we talk about headers, it is implicit they are HTTP headers. I would rather use something like:

- `BaseHTTPAuthHeaderBuilder implements HTTPAuthHeaderBuilder` 
- `HTTPAuthHeaderBuilder implements IHTTPAuthHeaderBuilder`

src/lib/clients/auth.ts
```diff
@@ -65,3 +65,51 @@ export class FakeAuthClient {
...
+}
+
+export class HTTPAuthHeaderBuilder implements AuthHeaderBuilder {
+  private gcpAuthClient?: GCPAuthClient | undefined;
+
+  constructor(gcpAuthClient?: GCPAuthClient) {
+    this.gcpAuthClient = gcpAuthClient;
+  }
+
+  async buildHeader(spotifyToken?: string): Promise<Record<string, string>> {
```
DanielMoraDC, line 80: I would rather keep this token abstract in case we change the platform. What about `appToken` or simply `token`?

DanielMoraDC, line 80: This token should always be provided, cannot be undefined.

src/lib/clients/auth.ts
```diff
@@ -65,3 +65,51 @@ export class FakeAuthClient {
...
+    if (!this.gcpAuthClient) return headers;
+
+    return {
+      ...headers,
+      [this.gcpAuthClient.getHeaderName()]: `Bearer ${await this.gcpAuthClient.getToken()}`,
+    };
+  }
+}
+
+export class FakeHTTPAuthHeaderBuilder implements AuthHeaderBuilder {
```
DanielMoraDC, line 94: Nitpick: This works but we can also keep it simpler by using a stub and let `buildHeader` simply returns headers that we pass in the constructor.

Resolve at will.

src/lib/clients/auth.test.ts
```diff
@@ -1,48 +1,116 @@
...
+  GCPHTTPAuthClient,
+  HTTPAuthHeaderBuilder,
+} from './auth';
 import { FakeHttpClient, HttpResponse, Method } from './http';
 
-describe('HTTPAuthClient', () => {
-  const audience = 'test-audience';
-  let httpClient: FakeHttpClient;
-  let response: HttpResponse;
+describe('AuthClient', () => {
```
DanielMoraDC, line 9: Nitpick: We are not acxtually testing the interface per se, just the implementations, so this is probably not accurate. Resolve at will.

src/lib/clients/auth.test.ts
```diff
@@ -1,48 +1,116 @@
...
+
+      await expect(authClient.getToken()).rejects.toThrow(errorMessage);
+    });
   });
 
-  it('should throw an error if the client fails', async () => {
-    const errorMessage = 'Test error';
-    httpClient.setSendErrorMessage('Test error');
-    const authClient = new GCPHTTPAuthClient(httpClient, audience);
+  describe('HTTPAuthHeaderBuilder', () => {
```
DanielMoraDC, line 55: Could we also apply the Given, When, Then separation in these tests? 🙇

src/lib/clients/auth.test.ts
```diff
@@ -59,45 +59,31 @@ describe('AuthClient', () => {
...
       vi.spyOn(authClient, 'getHeaderName');
       const client = new BaseHTTPAuthHeaderBuilder(authClient);
-      await client.buildHeader();
+      const token = 'app-token';
+      await client.buildHeader(token);
       expect(authClient.getToken).toHaveBeenCalled();
       expect(authClient.getHeaderName).toHaveBeenCalled();
     });
 
     it('should return the correct auth header if token is provided', async () => {
```
DanielMoraDC, line 73: I believe the name of the test (and the next one) should also be corrected accordingly.

src/pages/api/artists/search.ts
```diff
@@ -62,11 +61,20 @@ export function createSearchArtistHandler({
       return;
     }
 
+    const token = await getToken({ req: request });
+
+    if (!token) {
```
DanielMoraDC, line 66: Should we also test when the token is not provided?

src/lib/clients/artists.test.ts
```diff
@@ -1,10 +1,10 @@
 import { beforeEach, describe, it, expect, vi } from 'vitest';
 import { Artist } from '@/lib/artists';
-import { HTTPBackendClient } from './backend';
+import { ArtistsHTTPClient } from './artists';
 import { FakeHttpClient, HttpResponse, Method } from './http';
-import { FakeAuthClient } from './auth';
+import { FakeBaseHTTPAuthHeaderBuilder } from './auth';
 
-describe('HTTPBackendClient', () => {
+describe('ArtistsHTTPBackendClient', () => {
```
DanielMoraDC, line 7: Nitpick: Backend should be removed for consistency.

src/lib/clients/auth.ts
```diff
@@ -65,3 +65,51 @@ export class FakeAuthClient {
...
+    if (!this.gcpAuthClient) return headers;
+
+    return {
+      ...headers,
+      [this.gcpAuthClient.getHeaderName()]: `Bearer ${await this.gcpAuthClient.getToken()}`,
+    };
+  }
+}
+
+export class FakeBaseHTTPAuthHeaderBuilder implements HTTPAuthHeaderBuilder {
```
DanielMoraDC, line 94: I believe that we should be faking the interface, not the implementation. So it makes more sense to use a `HTTPAuthHeaderBuilderStub` that simply returns some header provided in the constructor.

src/lib/clients/auth.ts
```diff
@@ -5,7 +5,7 @@ export interface AuthClient {
   getHeaderName(): string;
 }
 
-export class GCPHTTPAuthClient {
+export class HTTPAuthClient {
```
DanielMoraDC, line 8: This class is GCP-focused, so I believe we should reflect that in the name (e.g. `GCPAuthClient`).

DanielMoraDC (approved):
Added the changes we discussed. LGTM

PR #25, opened 2025-02-16
feat: FW-57 retrieve existing playlists

barroro:
- Create `PlaylistsHTTPClient` with `searchPlaylists` method to retrieve all the existing playlists.
- Create a new API route to retrieve existing playlists using `PlaylistsHTTPClient`.
- Create the Playlist entity.


src/pages/api/playlists/search.ts
```diff
@@ -0,0 +1,111 @@
...
+    } catch {
+      response
+        .status(500)
+        .json({ message: 'Unexpected error, cannot retrieve playlists' });
+    }
+  };
+}
+
+const defaultLimit: number = parseInt(
+  process.env.SEARCH_ARTISTS_DEFAULT_LIMIT || '5'
```
DanielMoraDC, line 95:
```suggestion
  process.env.SEARCH_PLAYLIST_DEFAULT_LIMIT || '5'
```

src/pages/api/playlists/search.ts
```diff
@@ -0,0 +1,111 @@
...
+);
+const maxLimit: number = parseInt(process.env.SEARCH_ARTISTS_MAX_LIMIT || '10');
```
DanielMoraDC, line 97:
```suggestion
const maxLimit: number = parseInt(process.env.SEARCH_PLAYLIST_MAX_LIMIT || '10');
```

src/pages/api/playlists/search.ts
barroro: Thanks for reminding me! Done 👌

DanielMoraDC (approved):
LGTM, good job 🚀 

There is a bit of duplication but I would say that we abstract the clients whenever we see a third use case that can benefit from that, so we avoid adding too much complexity early on.

PR #26, opened 2025-03-07
feat: FW-74 display and select existing playlists in the generator form

barroro:
- Create a new context called ServiceProvider to provide services using dependency inversion.
- Create `SelectSearch` component using `Command` and `Popover` components.
- Create `FetchService` and `PlaylistService` to retrieve existing playlists from the UI.

### DIFF
--- a/src/hooks/useDebounceCallback.ts
+++ b/src/hooks/useDebounceCallback.ts
@@ -1,0 +1,26 @@
+import { useCallback, useRef } from 'react';
+
+/**
+ * Hook to debounce a callback function
+ * @param callback - Function to execute after the debounce
+ * @param delay - Delay in milliseconds
+ */
+export function useDebouncedCallback<T extends (..._args: any[]) => void>(
+  callback: T,
+  delay: number
+) {
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  return useCallback(
+    (...args: Parameters<T>) => {
+      if (timeoutRef.current) {
+        clearTimeout(timeoutRef.current);
+      }
+
+      timeoutRef.current = setTimeout(() => {
+        callback(...args);
+      }, delay);
+    },
+    [callback, delay]
+  );
+}
--- a/src/hooks/useDebounceCallback.ts
+++ b/src/hooks/useDebounceCallback.ts
@@ -7,7 +7,7 @@ import { useCallback, useRef } from 'react';
  */
 export function useDebouncedCallback<T extends (..._args: any[]) => void>(
   callback: T,
-  delay: number
+  delay: number = 500
 ) {
   const timeoutRef = useRef<NodeJS.Timeout | null>(null);
 
--- a/src/services/fetchService.ts
+++ b/src/services/fetchService.ts
@@ -1,0 +1,13 @@
+export interface IFetchService {
+  fetchData<T>(_url: string, _options?: RequestInit): Promise<T>;
+}
+
+export class FetchService implements IFetchService {
+  async fetchData<T>(url: string, options?: RequestInit): Promise<T> {
+    const response = await fetch(url, options);
+    if (!response.ok) {
+      throw new Error(`Error fetching data: ${response.statusText}`);
+    }
+    return response.json();
+  }
+}
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -8,6 +8,8 @@ import {
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import GeneratePlaylistPage, { GenerateProps } from '@/pages/generate';
+import { ReactNode } from 'react';
+import { ServiceProvider } from '@/contexts/ServiceContext';
 
 vi.mock('next/image', () => ({
   __esModule: true,
@@ -42,9 +44,23 @@ const clickToNextButton = async () => {
   await userEvent.click(nextButton);
 };
 
+const playlistsService = {
+  searchPlaylists: vi.fn(),
+};
+
+const MockServiceProvider = ({ children }: { children: ReactNode }) => {
+  return (
+    <ServiceProvider value={{ playlistsService }}>{children}</ServiceProvider>
+  );
+};
+
+const customRenderWithProviders = (ui: ReactNode) => {
+  return render(ui, { wrapper: MockServiceProvider });
+};
+
 describe('GeneratePlaylistPage', () => {
   it('should render the form with the first step displayed', async () => {
-    render(<GeneratePlaylistPage {...staticTranslations} />);
+    customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
 
     const firstNavigationStepButton = screen.getByRole('button', {
       name: /steps.step1.title/i,
@@ -74,7 +90,7 @@ describe('GeneratePlaylistPage', () => {
   });
 
   it('should display the error message when the "Create new playlist" option is selected and the playlist name is not filled', async () => {
-    render(<GeneratePlaylistPage {...staticTranslations} />);
+    customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
 
     const createNewPlaylistRadio = screen.getByRole('radio', {
       name: /steps.step1.form.createNewPlaylist.title/i,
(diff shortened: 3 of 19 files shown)


DanielMoraDC:
Given that there are 3 main changes, I suggest to create 3 separate PRs. It would make review process much lighter.

DanielMoraDC:
https://github.com/user-attachments/assets/a0f73492-2a7c-49de-8393-b4999a5bc5d3

I am testing the branch locally and I see the search is buggy. Results of the search only appears once I delete the text.

Also: could we use a text input directly? Having to click the dropdown adds an unnecessary extra step.

src/components/generate/PlaylistSetupForm/ExistingPlaylistsSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+  SelectSearchContent,
+} from '@/components/ui/SelectSearch';
+import { useDebouncedCallback } from '@/hooks/useDebounceCallback';
+import useTranslation from 'next-translate/useTranslation';
+import { useFormContext } from 'react-hook-form';
+import { useExistingPlaylists } from './useExistingPlaylists';
+import { useState } from 'react';
+import ErrorMessage from '@/components/ui/ErrorMessage';
+
+const getEmptyMessageKeyTranslation = (
```
DanielMoraDC, line 15: Nitpick: should this be `getEmptyMessageTranslationKey`, instead?

src/services/fetchService.ts
```diff
@@ -0,0 +1,13 @@
+export interface IFetchService {
```
DanielMoraDC, line 1: Shouldn't `services` be located inside `lib` along with the rest of the code?

DanielMoraDC, line 1: SHouldn't this go inside `lib`, too?

src/hooks/useDebounceCallback.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback, useRef } from 'react';
+
+/**
```
DanielMoraDC, line 3: Let's keep all comments in English for consistency 🙇 If this has been copied from some public snippet, I would also suggest to add the source.

src/components/generate/PlaylistSetupForm/ExistingPlaylistsSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  if (playlists.length === 0 && searchValue.trim() !== '') {
+    return 'steps.step1.form.useExistingPlaylist.playlistSelector.noResults';
+  }
+
+  return 'steps.step1.form.useExistingPlaylist.playlistSelector.searchPlaceholder';
+};
+
+const ExistingPlaylistsSelector = () => {
```
DanielMoraDC, line 31: I would rather name this `PlaylistSearcher`. The word `search` already tells us they must be existing ones and it is consistent with the use of `search` inside the context (instead of select).

src/components/generate/PlaylistSetupForm/useExistingPlaylists.ts
```diff
@@ -0,0 +1,32 @@
+import { useState, useCallback } from 'react';
+import { Playlist } from '@/services/playlistsService';
+import { useServices } from '@/contexts/ServiceContext';
+
+export function useExistingPlaylists() {
```
DanielMoraDC, line 5: I also think this name can be confusing. I understand keeping the `use` prefix for consistency, but what about `usePlaylistSearch`?

src/components/generate/PlaylistSetupForm/ExistingPlaylistsSelector.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [searchValue, setSearchValue] = useState('');
+  const { playlists, search, loading } = useExistingPlaylists();
+
+  const { control, formState } = useFormContext();
+  const { errors } = formState;
+
+  const debouncedSearch = useDebouncedCallback((searchTerm: string) => {
+    search(searchTerm);
+    setSearchValue(searchTerm);
+  }, 500);
```
DanielMoraDC, line 42: Can we keep this as a default parameter so we do not hardcode it?

src/components/ui/SelectSearch.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+export function SelectSearch({
+  value,
+  defaultValue,
+  onChange,
+  onSearch,
+  items,
+  children,
+  multiple = false,
```
DanielMoraDC, line 57: Can we just assume we are always going to receive a list, here, so we can keep this simpler?

src/components/ui/SelectSearch.tsx
```diff
@@ -0,0 +1,244 @@
...
+        updatedValue = currentValues.includes(newValue)
+          ? currentValues.filter((v) => v !== newValue)
+          : [...currentValues, newValue];
+      } else {
+        updatedValue = newValue;
+        setIsOpen(false);
+      }
+
+      setSelectedValue(updatedValue);
+      onChange?.(updatedValue);
```
DanielMoraDC, line 85: Shouldn't `onChange` always be provided?

src/components/ui/SelectSearch.tsx
```diff
@@ -0,0 +1,244 @@
...
+  value?: string | string[];
+  defaultValue?: string | string[];
+  onChange?: (_value: string | string[]) => void;
+  onSearch?: (_query: string) => Promise<void> | void;
+  items: SelectSearchItem[];
+  children: React.ReactNode;
+  multiple?: boolean;
+}
+
+export function SelectSearch({
```
DanielMoraDC, line 50: This part feels a bit confusing:
- We declare more props than the ones actually used.
- Some props are just passed so they are returned. Can't we just let the parent component handle that? Feels like an unnecessary burden for this component.
- We call this `SelectSearch` but its responsabilities are keeping the state of the selected value. Can we reflect that in the naming?

src/components/generate/PlaylistSetupForm/ExistingPlaylistsSelector.tsx
barroro: Yes, it's more consistent `PlaylistSearcher`. I'm going to change it. Good point!

src/components/ui/SelectSearch.tsx
barroro: What do you mean? If we can remove the `multiple` prop, right?

src/components/ui/SelectSearch.tsx
barroro: I've simplified the flow by removing the `SelectSearch` component because it was also causing confusion and complexity.

src/components/generate/PlaylistSetupForm/useExistingPlaylists.ts
barroro: Done! The `use` prefix is generally used in the hook functions to distinct between them.

src/services/fetchService.ts
barroro: I'm not sure. I've been looking for some NextJS structures to understand the different folders. However, I would like to separate the API and client methods. We could use `/lib` folder just for the API logic. What do you think?

DanielMoraDC:
Nice updates 🚀 

I see that the playlist is not properly stored in session when going back from the next step. Is that expected?


https://github.com/user-attachments/assets/5eaf045e-aae8-4e5e-b329-e2366efbbf24

src/services/fetchService.ts
barroro: Yes, we should have separated the API and the UI parts in different folders. I was thinking in how can we apply this approach but I didn't find the better solution right now. I think the best approach is using App router as I told you sometimes and probably I should spend time to apply this after the MVP. There are some parts in the `lib` folder that for me it makes sense to be in this folder like `utils`or the `motionVariants`. I would discuss this in another PR, we can create specific task to refactor this part to avoid create more changes in this MR. What do you think?

DanielMoraDC: Sure, this is not something introduced in this PR, so happy to move the conversation outside. Resolving.

DanielMoraDC (approved):

PR #27, opened 2025-03-22
refactor(entities): create entities folder for the Classes and DTOs

barroro:


src/pages/api/artists/search.ts
```diff
@@ -12,14 +12,9 @@ export type SearchArtistHandlerParams = {
...
 
-type ResponseArtist = {
-  name: string;
-  imageUri: string | undefined;
-};
-
 type ResponseData = {
   message: string;
-  artists?: ResponseArtist[];
+  artists?: Array<ArtistDTO>;
```
DanielMoraDC, line 17: I am good with using a more standard name `DTO`, but should we keep that in this layer, close to where it is used?

src/entities/playlists.ts
```diff
@@ -1,3 +1,33 @@
+/* eslint-disable no-unused-vars */
+export enum GeneratePlaylistType {
```
DanielMoraDC, line 2: I would keep names decoupled to how they are used. So I would rather name this `PlaylistType`, instead.

src/entities/playlists.ts
```diff
@@ -1,3 +1,33 @@
...
+  New = 'new',
+  Existing = 'existing',
+}
+
+export type GenerateNewPlaylist = {
```
DanielMoraDC, line 7: Same for `GenerateNewPlaylist` and `GenerateExistingPlaylist`. Moreover, using verbs as prefix can be confusing for data types, as they are usually more related to methods/functions. I feel can be more concise:

```ts
export type NewPlaylist = {
  type: PlaylistType.New;
  name: string;
  isPrivate: boolean;
  bands: string[];
};

export type ExistingPlaylist = {
  type: PlaylistType.Existing;
  id: string;
  bands: string[];
};
```

src/entities/playlists.ts
```diff
@@ -1,3 +1,33 @@
...
+  playlistType: GeneratePlaylistType.New;
+  name: string;
+  isPrivate: boolean;
+  bands: string[];
```
DanielMoraDC, line 11: We are using both `bands` and `artists` in different places, which can be confusing. I already created a ticket for this, just sharing it here for visibility. Resolving!

src/pages/api/artists/search.ts
DanielMoraDC: Nevermind, I think it is good like this. Resolving!

src/entities/playlists.ts
```diff
@@ -1,3 +1,33 @@
...
+};
+
+export type GeneratePlaylistDTO =
+  | GenerateNewPlaylist
+  | GenerateExistingPlaylist;
+
+export type PlaylistDTO = {
+  id: string;
+  name: string;
+  description: string | undefined;
```
DanielMoraDC, line 27: For consistency, shouldn't we keep `description?`, instead?

barroro: Done!

src/entities/playlists.ts
barroro: I've changed the names to be more descriptive and I've removed the type prop from the DTOs, because it doesn't make sense. This value is only to manage which DTO and request we have to use depending on the selection in the form.

src/entities/playlists.ts
barroro: I've changed the name for `PlaylistCreationMode`. It's more descriptive than the previous one.

DanielMoraDC (approved):

PR #28, opened 2025-03-23
feat: FW-70 display artists in the form retrieving from the backend

barroro:


src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
```diff
@@ -69,9 +54,13 @@ const PlaylistSearchBandsForm = () => {
             <FormItem>
               <FormControl>
                 <SearchBandsCombobox
-                  options={options}
+                  options={artists}
                   values={field.value}
-                  onChange={field.onChange}
+                  onChange={(value) => {
+                    console.log({ value });
```
DanielMoraDC, line 60: Let's make sure to delete this.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -1,21 +1,21 @@
 import { CircleCheck } from 'lucide-react';
-import Image, { StaticImageData } from 'next/image';
+import Image from 'next/image';
 
 interface BandSearchResultProps {
```
DanielMoraDC, line 4: Maybe this is also a good chance to make this a `type`, instead, for consistency.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -24,13 +24,13 @@ const BandSearchResult = ({
...
+      {srcImage ? (
         <Image
-          src={icon.src}
-          alt=""
-          width={icon.height}
-          height={icon.width}
-          className="h-8 w-8 rounded-md object-cover mr-2"
+          src={srcImage}
+          alt={name}
+          width={30}
```
DanielMoraDC, line 31: Should we move this into a prop with a default value? So it is easier to modify later.

src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
```diff
@@ -1,55 +1,40 @@
...
 
 const PlaylistSearchBandsForm = () => {
   const { control, watch, setValue, formState } = useFormContext();
+  const { artists, search } = useArtistSearch();
   const { errors } = formState;
 
   const { t } = useTranslation('generate');
 
-  const selectedValues: Array<number> = watch('bands', []);
+  const selectedValues: Array<string> = watch('bands', []);
```
DanielMoraDC, line 21: For better readability, I suggest we use more concise nambes for these functions/variables. For example:

- `setValue` -> `setSelectedBands`.
- `removeSelectedItems` -> `removeSelectedBand`.
- `onChangeSelection` -> `onBandSelectionChange`.
- `selectedValues` -> `selectedBands`.

src/components/generate/PlaylistSearchBandsForm/useArtistSearch.ts
```diff
@@ -0,0 +1,32 @@
...
+  const search = useCallback(
+    async (name: string, limit: number = 5) => {
+      try {
+        if (name.trim() === '') {
+          setArtists([]);
+          return;
+        }
+        setLoading(true);
+        const data = await artistsService.searchArtists(name, limit);
+        console.log(data);
```
DanielMoraDC, line 20: Let's remember to also remove this one 🕵️

DanielMoraDC:
Good job! Just found a small thing: after selecting an artist, the list is not cleared and probably should.

https://github.com/user-attachments/assets/b6fc5fd8-a5df-4a04-aef2-88fa390d7923

Also, remember to add a description to the PR when ready 😃

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
barroro: I used this type for the mock data because I wasn't using URL strings. For the current scenario, we don't need this type because we are using strings.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
barroro: I would use it as a static value, like a constant because it shouldn't change regularly. What do you think?

src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
barroro: Good point! I'm totally agree. I put these names at the beginning because I wanted to create a generic component but at the end it didn't make sense. I'm going to change it!

src/components/generate/PlaylistSearchBandsForm/useArtistSearch.ts
barroro: done!

src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
barroro: I was thinking to unify the name and use `artist` instead of `band`. What do you think? The idea would be to change the file names as well.

src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
barroro: Done!

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
DanielMoraDC: Not sure we are talking about the same thing 😅 

I mean using `type` for defining `Props` instead of using interface.

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
DanielMoraDC: Constant somewhere definitely better than hardcoded. I would personally prefer using an environment variable to read the value and using the constant as fallback. By doing so, if we need to change the value we do not require a full image build.

src/components/generate/PlaylistSearchBandsForm/PlaylistSearchBandsForm.tsx
DanielMoraDC: I definitely want that to happen 🙇  I would recommend doing that in a different PR, though, as it probably requires a lot of changes in different files.

barroro: It's true. I'll do it after this PR 👌

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
barroro: Done!

__tests__/generate.test.tsx
```diff
@@ -44,14 +44,76 @@ const clickToNextButton = async () => {
   await userEvent.click(nextButton);
 };
 
+const completeFirstStepNewPlaylist = async () => {
```
DanielMoraDC, line 47: Following the Arrange-Act-Assert [pattern](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/), I think it is convenient that our auxiliar functions only belong to one of those steps. In these ones, we see how they do some arrangements, then act, then do expectations. In some sense, they are already a test by themselves 😅

__tests__/generate.test.tsx
```diff
@@ -255,23 +315,42 @@ describe('GeneratePlaylistPage', () => {
     });
     await userEvent.click(useExistingPlaylistRadio);
 
-    const triggerInput = screen.getByRole('combobox');
-    await userEvent.click(triggerInput);
+    await findAndSelectPlaylist('My playlist');
+  });
+
+  it('should add multiple artists and remove some of them when clicking the remove button', async () => {
```
DanielMoraDC, line 321: Isn't this testing too many behaviors? Following the idea in [this other comment](https://github.com/DanielMoraDC/festwrap-ui/pull/28/files#r2019070140), I would separate them into multiple ones to make it clear and explicit what is being tested. For example:

- "should show the artist search screen on selecting the new playlist mode"
- "should display searched artists on input change"
- "should display artist as selected when clicking on a search"
- "should remove selected artist on remove click"

What do you think?

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
DanielMoraDC: Any opinion on this ⬆️ ?

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -1,21 +1,23 @@
...
+const IMAGE_SIZE = 30;
 
 interface BandSearchResultProps {
   key: number;
   name: string;
   isActive: boolean;
   isSelected: boolean;
   handleItemSelect: (_item: any) => void;
-  icon?: StaticImageData;
+  srcImage?: string;
```
DanielMoraDC, line 12: Nitpick: `imageUrl` probably is more concise than `srcImage`

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
barroro: I thought that you were talking about another thing. Yes, it should be a `type`. I'm going to change it!

barroro: Done!

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
```diff
@@ -1,21 +1,23 @@
 import { CircleCheck } from 'lucide-react';
-import Image, { StaticImageData } from 'next/image';
+import Image from 'next/image';
+
+const IMAGE_SIZE = 30;
```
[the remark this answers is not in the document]
barroro: I don't usually use `env` variables for this kind of thing. If it belongs to the UI layer, I usually use constants. But it might make sense.

__tests__/generate.test.tsx
barroro: Yes, it's true but I thought that you need to do the previous steps to do some checks and I thought to create a test like a user scenario to avoid repeating some steps multiple times in the tests.  What do you think?

__tests__/generate.test.tsx
barroro: True, good point! I'll change it to move the expect to the test case. 👌

barroro: Done!

DanielMoraDC: I still believe it should not ideally be there. If the element to check is defined inside, we can also move it outside like this:

```ts

const  secondStepContentTitle = await waitFor(() => {
    const secondStepContent = screen.getByRole('tabpanel');
    return within(secondStepContent).getByText(/steps.step2.title/i);
  });

const completeFirstStepNewPlaylist = async () => {
  const playlistNameInput = screen.getByLabelText(
    /steps.step1.form.createNewPlaylist.giveAName/i
  );
  await userEvent.type(playlistNameInput, 'My new playlist');

  await clickToNextButton();
};

[...]
await completeFirstStepNewPlaylist()

expect (await secondStepContentTitle()).toBeInTheDocument()
```

Does it make sense?

src/components/generate/PlaylistSearchBandsForm/BandSearchResult.tsx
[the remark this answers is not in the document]
DanielMoraDC: I am reading that these type of variables cannot be dynamically set after build ([docs](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser)), so it does not make sense to do that. Resolving ✅

__tests__/generate.test.tsx
DanielMoraDC: Given that you have auxiliar functions for the test preparations, they should not add much overhead, I believe. The problem with such tests is that it is not clear exactly what is being tested and it can fail in multiple points, which makes it harder to debug. Moreover, if we keep adding complexity to that screen this test will become larger, which is inconvinient.

I am fine resolving this as long as we create a separate issue to tackle it.

__tests__/generate.test.tsx
barroro: Yes, it makes sense! I'm going to change it.

__tests__/generate.test.tsx
```diff
@@ -255,23 +310,86 @@ describe('GeneratePlaylistPage', () => {
...
-    const searchInput = screen.getByPlaceholderText(
-      /steps.step1.form.useExistingPlaylist.playlistSelector.placeholder/i
-    );
-    await userEvent.type(searchInput, 'Holding');
+    const selectedItem = screen.getByText(playlistName);
+    expect(selectedItem).toBeInTheDocument();
+  });
+
+  it('should select multiple artists when searching for them and clicking on them', async () => {
+    artistsService.searchArtists.mockResolvedValue({
```
DanielMoraDC, line 321: Would it be a good idea to mock this by default? Would probably make test shorter and prevent duplicated code in the next test.

DanielMoraDC: Resolve at will.

barroro: I prefer to have more control over this part. In my opinion, is more semantic and more descriptive work on this way.

__tests__/generate.test.tsx
```diff
@@ -255,23 +310,86 @@ describe('GeneratePlaylistPage', () => {
...
+
+    await waitFor(async () => {
+      const removeButton = screen.getByLabelText(
+        'steps.step2.removeArtist Holding Absence'
+      );
+      expect(removeButton).toBeInTheDocument();
+      await userEvent.click(removeButton);
+    });
+
+    const selectedItemRemoved = screen.queryByText('HOLD');
```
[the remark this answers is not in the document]
barroro: Sorry, it's true. It's already fixed 🙏

__tests__/generate.test.tsx
```diff
@@ -255,23 +310,86 @@ describe('GeneratePlaylistPage', () => {
...
+
+    await waitFor(async () => {
+      const removeButton = screen.getByLabelText(
+        'steps.step2.removeArtist Holding Absence'
+      );
+      expect(removeButton).toBeInTheDocument();
+      await userEvent.click(removeButton);
+    });
+
+    const selectedItemRemoved = screen.queryByText('Holding Absence');
```
DanielMoraDC, line 392: Given you already have the artist in `artistName`, we could probably reuse it here and in line 386.

DanielMoraDC (approved):
LGTM

Left a final comment, feel free to self-resolve it and merge.

__tests__/generate.test.tsx
barroro: Done!

PR #30, opened 2025-04-04
feat: [FW-86] Create new POST endpoint in API to create a new playlist

barroro:
# Description
Implement a new API endpoint to create a new playlist.
- Add handling errors
- Add test cases for API endpoint and new method in playlist client

# Testing
- You have to take the cookie from the request using the client app and executing for any GET endpoint from the request in Chrome Devtools.

image

```
curl --location 'http://localhost:3000/api/playlists/create' \
--header 'Cookie: <NEXT-AUTH-TOKEN>' \
--header 'Content-Type: application/json' \
--data '{
    "playlist": {
        "name": "Test playlist 4",
        "description": "Test description",
        "isPublic": false,
        "artists": [
            { "name": "Holding Absence" },
            { "name": "Attila" },
            { "name": "Polaris" },
            { "name": "Parkway Drive" }
        ]
    }
}'
```

### DIFF
--- a/src/entities/playlists.ts
+++ b/src/entities/playlists.ts
@@ -1,3 +1,5 @@
+/* eslint-disable no-unused-vars */
+
 export type CreateNewPlaylistDTO = {
   playlist: {
     name: string;
@@ -9,13 +11,14 @@ export type CreateNewPlaylistDTO = {
   }>;
 };
 
-export type CreateNewPlaylistStatusType =
-  | 'CREATED_WITHOUT_ISSUES'
-  | 'CREATED_MISSING_ARTISTS';
+export enum CreatedPlaylistStatus {
+  OK = 'OK',
+  MISSING_ARTISTS = 'MISSING_ARTISTS',
+}
 
 export type CreateNewPlaylistResponseDTO = {
   id: string;
-  status: CreateNewPlaylistStatusType;
+  status: CreatedPlaylistStatus;
 };
 
 export type UpdatePlaylistDTO = {
--- a/src/entities/playlists.ts
+++ b/src/entities/playlists.ts
@@ -9,8 +9,13 @@ export type CreateNewPlaylistDTO = {
   }>;
 };
 
+export type CreateNewPlaylistStatusType =
+  | 'CREATED_WITHOUT_ISSUES'
+  | 'CREATED_MISSING_ARTISTS';
+
 export type CreateNewPlaylistResponseDTO = {
   id: string;
+  status: CreateNewPlaylistStatusType;
 };
 
 export type UpdatePlaylistDTO = {
--- a/src/entities/playlists.ts
+++ b/src/entities/playlists.ts
@@ -4,7 +4,7 @@ export type CreateNewPlaylistDTO = {
   playlist: {
     name: string;
     description?: string | undefined;
-    isPrivate: boolean;
+    isPublic: boolean;
   };
   artists: Array<{
     name: string;
--- a/src/entities/playlists.ts
+++ b/src/entities/playlists.ts
@@ -1,7 +1,16 @@
 export type CreateNewPlaylistDTO = {
-  name: string;
-  isPrivate: boolean;
-  artists: string[];
+  playlist: {
+    name: string;
+    description?: string | undefined;
+    isPrivate: boolean;
+  };
+  artists: Array<{
+    name: string;
+  }>;
+};
+
+export type CreateNewPlaylistResponseDTO = {
+  id: string;
 };
 
 export type UpdatePlaylistDTO = {
--- a/src/lib/clients/playlists.test.ts
+++ b/src/lib/clients/playlists.test.ts
@@ -124,19 +124,23 @@ describe('PlaylistsHTTPClient', () => {
         },
         artists: [{ name: 'Artist 1' }, { name: 'Artist 2' }],
       };
+      authHeaderBuilder = new AuthHeaderBuilderStub();
+    });
+
+    const mockHttpClientResponse = (status: number = 201) => {
       response = {
         data: {
           playlist: {
             id: '123',
           },
         },
-        status: 201,
+        status,
       };
       httpClient = new FakeHttpClient(response);
-      authHeaderBuilder = new AuthHeaderBuilderStub();
-    });
+    };
 
     it('should call the client with the correct parameters', async () => {
+      mockHttpClientResponse();
       const headers = { authorization: 'Bearer token' };
       const client = new PlaylistsHTTPClient(
         url,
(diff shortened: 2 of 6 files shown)


src/entities/playlists.ts
```diff
@@ -1,7 +1,16 @@
...
+    description?: string | undefined;
+    isPrivate: boolean;
+  };
+  artists: Array<{
+    name: string;
+  }>;
+};
+
+export type CreateNewPlaylistResponseDTO = {
+  id: string;
```
DanielMoraDC, line 13: The server returns:
- 201 if the playlist has been created without issues.
- 207 if the playlist has been created but some artists could not be added.

Should we accomodate the response to also reflect that (maybe we want to add some warning message for the case when not all artists could be properly added).

What do you think?

barroro: Cool! I didn't understand the reason for the 207 status. I'm going to implement a message based on the status 👌

barroro: Done! I'm returning a custom status message in the API response to get more info about the request in the UI. What do you think?

src/pages/api/playlists/create.ts
```diff
@@ -67,7 +67,15 @@ export function createCreatePlaylistHandler({
...
-      response.status(200).json({
+      if (playlistCreated.status === 'CREATED_MISSING_ARTISTS') {
+        response.status(207).json({
+          playlistCreated,
+          message:
+            'Playlist has been created but some artists could not be added',
+        });
+      }
+
+      response.status(201).json({
```
DanielMoraDC, line 78: Shouldn't this be in an else?

src/entities/playlists.ts
DanielMoraDC: Would an enum be more suitable? Something like this:

```typescript
enum CreatedPlaylistStatus {
  OK = "ok",
  MISSING_ARTISTS = "artists missing"
}
```

I remove the `CREATED` suffix, as it is already implified.

src/lib/clients/playlists.test.ts
```diff
@@ -1,91 +1,227 @@
...
+        method: Method.Post,
+        data: playlistData,
+        headers: headers,
+      });
     });
-  });
 
-  it('should return the list of playlists returned by the HTTP client', async () => {
-    const client = new PlaylistsHTTPClient(url, httpClient, authHeaderBuilder);
+    it('should return the response data from the HTTP client without issues when the status is 201', async () => {
```
DanielMoraDC, line 162: Nitpick: given this and the next are very similar, we could parametrize the test a keep only one ([example](https://www.the-koi.com/projects/parameterized-data-driven-tests-in-vitest-example/)). Resolve at will.

src/lib/clients/playlists.ts
```diff
@@ -50,16 +58,54 @@ export class PlaylistsHTTPClient implements PlaylistsClient {
...
+    const authHeader = await this.httpAuthHeaderBuilder.buildHeader(token);
+    return this.httpClient
+      .send({
+        url: `${this.url}/playlists`,
+        method: Method.Post,
+        data: playlist,
+        headers: authHeader,
+      })
+      .then((response: any) => {
+        if (response.status === 201) {
```
DanielMoraDC, line 75: Sould we fail if the status is any other than the expected ones (e.g. 500)?

DanielMoraDC: I think we also missed this in the rest of the clients. We can fix the others in a separate PR, though.

src/pages/api/playlists/create.ts
barroro: It could be, but it should return one of these two status and from the playlist client is managing the status to return of of these. There's no more status that we have manage if it works, right?

src/entities/playlists.ts
barroro: Good point! It's cleaner than the current approach. I'm going to change it!

src/lib/clients/playlists.test.ts
barroro: True, it makes sense. I'm going to change it 👌

src/lib/clients/playlists.ts
barroro: True, yesterday I saw the same problem but I though that the try/catch is in the API method but it always return `500`. But I'm agree with you, we could fix in a separate PR. I will create a task to fix it 👌

barroro: Done! I've created a task to manage, [FW-89](https://festwrap.atlassian.net/browse/FW-89)

src/entities/playlists.ts
barroro: Done!

src/lib/clients/playlists.test.ts
barroro: Done!

src/lib/clients/playlists.ts
DanielMoraDC: I am happy to fix the other API endpoints in a separate PR, but I believe we should include the error handling for creating a new playlist in this one, so the change we introduce is not buggy. What do you think?

src/pages/api/playlists/create.ts
DanielMoraDC: The main problem is that the code does not return on `response.status` or `response.json`, so we are actually setting the response twice every time. The test pass because we are checking that `response.json` and `response.status` calls exists, but it does not check whether more than one exists. I am not sure how NextJS handles this, but if the last call is the one being returned, this implementation would not work.

I would suggest we do not depend on the NextJS implementation specifics and make the behavior explicit: handle each case in separate branch conditions and add a (regression) check in the test to check that each call is only performed twice.

barroro: Done! I put a return inside the `if' condition to make sure the call is made once.

src/pages/api/playlists/create.ts
```diff
@@ -0,0 +1,107 @@
...
+        });
+        return;
+      }
+
+      response.status(201).json({
+        playlistCreated,
+        message: 'Playlists successfully created',
+      });
+    } catch {
+      response
```
DanielMoraDC, line 87: The only code likely to return an error is the playlist creation, so I would rather wrap the `createPlaylist` call, only.

src/lib/clients/playlists.ts
barroro: Perfect! Make sense. We can merge the other PR before.

src/pages/api/playlists/create.ts
barroro: I've moved the `playlistData` out of the try block 👌

DanielMoraDC (approved):

PR #31, opened 2025-04-06
fix: FW-89 Return error on 500 or unknown status

DanielMoraDC:
# Description

Clients to raise an error if the response status differs from the expected successful one.

### DIFF
--- a/src/lib/clients/artists.test.ts
+++ b/src/lib/clients/artists.test.ts
@@ -83,7 +83,7 @@ describe('ArtistsHTTPClient', () => {
     { status: 400, message: 'Bad request' },
     { status: 500, message: 'Internal error' },
   ])(
-    'should throw an error if the server response status it not the expected successful one',
+    'should throw an error if the server response status is not the expected successful one',
     async ({ status, message }) => {
       response = { status: status, data: message };
       httpClient = new FakeHttpClient(response);
--- a/src/lib/clients/artists.test.ts
+++ b/src/lib/clients/artists.test.ts
@@ -78,4 +78,21 @@ describe('ArtistsHTTPClient', () => {
 
     await expect(client.searchArtists(token, name, limit)).rejects.toThrow();
   });
+
+  it.each([
+    { status: 400, message: 'Bad request' },
+    { status: 500, message: 'Internal error' },
+  ])(
+    'should throw an error if the server response status it not the expected successful one',
+    async ({ status, message }) => {
+      response = { status: status, data: message };
+      httpClient = new FakeHttpClient(response);
+      const client = new ArtistsHTTPClient(url, httpClient, authHeaderBuilder);
+
+      const expectedMessage = `Unexpected artist search response status: ${status}: ${message}`;
+      await expect(client.searchArtists(token, name, limit)).rejects.toThrow(
+        expectedMessage
+      );
+    }
+  );
 });
--- a/src/lib/clients/artists.ts
+++ b/src/lib/clients/artists.ts
@@ -38,11 +38,17 @@ export class ArtistsHTTPClient implements ArtistsClient {
         params: { name, limit },
         headers: authHeader,
       })
-      .then((response) =>
-        response.data.map(
-          (artist: any) => new Artist(artist.name, artist.imageUri)
-        )
-      );
+      .then((response) => {
+        if (response.status == 200) {
+          return response.data.map(
+            (artist: any) => new Artist(artist.name, artist.imageUri)
+          );
+        } else {
+          throw new Error(
+            `Unexpected artist search response status: ${response.status}: ${response.data}`
+          );
+        }
+      });
   }
 }
 
--- a/src/lib/clients/artists.ts
+++ b/src/lib/clients/artists.ts
@@ -39,7 +39,7 @@ export class ArtistsHTTPClient implements ArtistsClient {
         headers: authHeader,
       })
       .then((response) => {
-        if (response.status == 200) {
+        if (response.status === 200) {
           return response.data.map(
             (artist: any) => new Artist(artist.name, artist.imageUri)
           );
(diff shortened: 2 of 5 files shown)


src/lib/clients/artists.ts
```diff
@@ -38,11 +38,17 @@ export class ArtistsHTTPClient implements ArtistsClient {
...
-          (artist: any) => new Artist(artist.name, artist.imageUri)
-        )
-      );
+      .then((response) => {
+        if (response.status === 200) {
+          return response.data.map(
+            (artist: any) => new Artist(artist.name, artist.imageUri)
+          );
+        } else {
+          throw new Error(
```
barroro, line 47: I think that we need to add the catch part, if the endpoint throw a `500` status, it shouldn't enter in this part, it should enter in the `catch` part.
I prefer to use async/await approach with try/catch to use the same approach as API endpoint. What do you think?

src/lib/clients/playlists.test.ts
```diff
@@ -88,4 +88,25 @@ describe('PlaylistsHTTPClient', () => {
...
+      response = { status: status, data: message };
+      httpClient = new FakeHttpClient(response);
+      const client = new PlaylistsHTTPClient(
+        url,
+        httpClient,
+        authHeaderBuilder
+      );
+
+      const expectedMessage = `Unexpected playlist search response status: ${status}: ${message}`;
+      await expect(client.searchPlaylists(token, name, limit)).rejects.toThrow(
```
barroro, line 107: You are not rejecting the fetch call, it works for this reason I think. We should reject the `send` method.

src/lib/clients/artists.test.ts
```diff
@@ -78,4 +78,21 @@ describe('ArtistsHTTPClient', () => {
...
+    { status: 500, message: 'Internal error' },
+  ])(
+    'should throw an error if the server response status is not the expected successful one',
+    async ({ status, message }) => {
+      response = { status: status, data: message };
+      httpClient = new FakeHttpClient(response);
+      const client = new ArtistsHTTPClient(url, httpClient, authHeaderBuilder);
+
+      const expectedMessage = `Unexpected artist search response status: ${status}: ${message}`;
+      await expect(client.searchArtists(token, name, limit)).rejects.toThrow(
```
barroro, line 93: You are not rejecting the fetch call, it works for this reason I think. We should reject the `send`method.

DanielMoraDC: I am not sure I understand what you mean. Could you elaborate?

src/lib/clients/artists.ts
DanielMoraDC: To implement that, we should make `send` throw an error on non-2xx status. I do not like to add too extra complexity to such class or make additional assumptions (e.g. what if the expected behavior for some code is not a 2xx?). It should be up to their clients to decide when to raise an exception. In any case, if we decide to add such behavior to the client, I would rather add an additional method `sendOrThrow` that actually does that so we keep the class flexible.

In any case, what is the advantage of adding a try/catch, here?

src/lib/clients/artists.test.ts
barroro: Sorry, I didn't explain very well. I think that the problem is you are not reproducing the right flow, because you are only reviewing the status but to reproduce the real scenario, you should have the `send` method mock and throw the error from this part. Did you test the functionality? Because I think that the throw should enter the `catch` block instead of the `then` block. Chek docu: [link](https://developer.mozilla.org/es/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch#ejemplos)

DanielMoraDC: I am using a `FakeClient` test double that returns a custom status code and message (line 89). On receiveing those non-successful status, the `searchArtists` code raises an error, which is the thing we are checking here. The `send` does not return any error depending on the status code.

src/lib/clients/artists.ts
barroro: There's no a big difference between the two approaches. For me the main reason is to unify the same approach between the clients and the API endpoint. Right now we are using promise callbacks in the clients and `async/await` in the API methods.
If you try to reproduce an exception, you will see that it's not enter in the `then` callback, it will enter in the `catch` method that is not defined yet.

Check this example:
- Put this code in the search artists method:
```
.then((response) => {
        if (response.status === 200) {
          return response.data.map(
            (artist: any) => new Artist(artist.name, artist.imageUri)
          );
        } else {
          console.error(
            `Unexpected artist search response status: ${response.status}: ${response.data}`
          );
          throw new Error(
            `Unexpected artist search response status: ${response.status}: ${response.data}`
          );
        }
      })
      .catch((error) => {
        console.error(`Error searching artists: ${error}`);
        throw new Error(`Error searching artists: ${error}`);
      });
```
- Then stop the server
- Try to search an artist and check the logs in the UI terminal

DanielMoraDC: > There's no a big difference between the two approaches. For me the main reason is to unify the same approach between the clients and the API endpoint. Right now we are using promise callbacks in the clients and async/await in the API methods.

It's true that we are using await and then mixed in the code. Should we stick to one? I think `await` makes code more readable, overall. What do you think? We can create a task to unify the style.

> If you try to reproduce an exception, you will see that it's not enter in the then callback, it will enter in the catch method that is not defined yet.

I am not sure I am following. If the server does not exist, `send` will propagate an error to the code that uses it. In our case, this is the API handler, which would catch the error ([here](https://github.com/festwrap/festwrap-ui/blob/main/src/pages/api/artists/search.ts#L82)).

barroro: It works right now! I'm not sure if using the `then` callback is the best way to handle the errors, but we can check later.

barroro (approved):
Good job! 👌

PR #32, opened 2025-04-09
chore: FW-90 use async/await consistency

DanielMoraDC:
# Description

Enforces using async/await consistently instead of `then`, as agreed, to improve readability.


barroro (approved):
Good job!

PR #33, opened 2025-04-13
feat: [FW-87] submit generator form to create a playlist from the new playlist option

barroro:
## Description
Implement all the flow to submit the form calling the new API endpoint to create a new playlist from the "New playlist" option.
- Add notifications for the feedback errors.
- Define a new method inside `playlistService` to create the new playlist.
- Replace `isPrivate` field for the `isPublic` field instead.
- Cover with tests.

## Testing
- Run `npm i` to install the new dependencies

### DIFF
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -250,7 +250,79 @@ describe('GeneratePlaylistPage', () => {
     });
   });
 
-  it('should generate the new playlist and copy the URL link into the clipboard', async () => {
+  it('should create the new playlist and display the success message', async () => {
+    artistsService.searchArtists.mockResolvedValue({
+      artists: [
+        {
+          name: 'Holding Absence',
+          imageUri: null,
+        },
+      ],
+    });
+    playlistsService.createNewPlaylist.mockResolvedValue({
+      playlistCreated: {
+        id: '123',
+      },
+    });
+
+    customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
+
+    const playlistNameInput = screen.getByLabelText(
+      /steps.step1.form.createNewPlaylist.giveAName/i
+    );
+    await userEvent.type(playlistNameInput, 'My new playlist');
+
+    const playlistDescriptionInput = screen.getByLabelText(
+      /steps.step1.form.createNewPlaylist.giveADescription/i
+    );
+    await userEvent.type(
+      playlistDescriptionInput,
+      'My new playlist description'
+    );
+
+    await clickToNextButton();
+
+    const secondStepContentTitle = await waitFor(() => {
+      const secondStepContent = screen.getByRole('tabpanel');
+      return within(secondStepContent).getByText(/steps.step2.title/i);
+    });
+
+    expect(secondStepContentTitle).toBeInTheDocument();
+
+    await findAndSelectArtist('Holding Absence');
+
+    const generateButton = screen.getByRole('button', {
+      name: /steps.navigation.generate/i,
+    });
+    await userEvent.click(generateButton);
+
+    const thirdStepContentTitle = await waitFor(() => {
+      const thirdStepContent = screen.getByRole('tabpanel');
+      return within(thirdStepContent).getByText(/steps.step3.title/i);
+    });
+
+    expect(thirdStepContentTitle).toBeInTheDocument();
+
+    const successfullyMessage = screen.getByText(
+      /steps.step3.playlisyGeneratedSuccessfully/i
+    );
+    expect(successfullyMessage).toBeInTheDocument();
+
+    expect(playlistsService.createNewPlaylist).toHaveBeenCalledWith({
+      playlist: {
+        name: 'My new playlist',
+        description: 'My new playlist description',
+        isPublic: true,
+      },
+      artists: [
+        {
+          name: 'Holding Absence',
+        },
+      ],
+    });
+  });
+
+  it('should copy the URL link into the clipboard when clicking the copy button', async () => {
     const spyClipboardWriteText = vi
       .spyOn(navigator.clipboard, 'writeText')
       .mockImplementation(() => Promise.resolve());
@@ -298,11 +370,6 @@ describe('GeneratePlaylistPage', () => {
     });
 
     expect(thirdStepContentTitle).toBeInTheDocument();
-
-    const successfullyMessage = screen.getByText(
-      /steps.step3.playlisyGeneratedSuccessfully/i
-    );
-    expect(successfullyMessage).toBeInTheDocument();
 
     const copyURLButton = screen.getByRole('button', {
       name: /steps.step3.copyButton/i,
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -10,6 +10,7 @@ import userEvent from '@testing-library/user-event';
 import GeneratePlaylistPage, { GenerateProps } from '@/pages/generate';
 import { ReactNode } from 'react';
 import { ServiceProvider } from '@/contexts/ServiceContext';
+import { toast } from 'sonner';
 
 vi.mock('next/image', () => ({
   __esModule: true,
@@ -17,6 +18,13 @@ vi.mock('next/image', () => ({
     const { src, alt } = props;
     // eslint-disable-next-line @next/next/no-img-element
     return <img src={src} alt={alt} {...props} />;
+  },
+}));
+
+vi.mock('sonner', () => ({
+  toast: {
+    success: vi.fn(),
+    error: vi.fn(),
   },
 }));
 
@@ -96,6 +104,7 @@ const findAndSelectPlaylist = async (playlistName: string) => {
 
 const playlistsService = {
   searchPlaylists: vi.fn(),
+  createNewPlaylist: vi.fn(),
 };
 
 const artistsService = {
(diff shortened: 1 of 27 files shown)


src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -0,0 +1,79 @@
...
+        const response =
+          await playlistsService.createNewPlaylist(newPlaylistData);
+
+        const { playlistCreated } = response;
+
+        return {
+          success: true,
+          data: playlistCreated?.id,
+        };
+      } else {
```
DanielMoraDC, line 55: I would suggest we actually check whether the is `PlaylistCreationMode.Existing` and then add a default error condition when the value is neither of them. I believe it will help debugging potential future issues.

src/pages/api/playlists/create.ts
```diff
@@ -36,9 +36,11 @@ export function createCreatePlaylistHandler({
 }: CreatePlaylistHandlerParams) {
   return async function handler(
     request: NextApiRequest,
-    response: NextApiResponse<ResponseData>
+    response: NextApiResponse<CreatePlaylistResponseData>
   ): Promise<void> {
-    const parsedArgs = searchQuerySchema.safeParse(request.body);
+    const body = JSON.parse(request.body);
+    const parsedArgs = searchQuerySchema.safeParse(body);
```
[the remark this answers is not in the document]
barroro: True, it was my fault to copy from another place. I'm going to change it 👌

src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const submitPlaylist = async (
+    values: FormSchemaType
+  ): Promise<SubmitPlaylistResponse> => {
+    const { playlistCreationMode, ...playlistData } = values;
+
+    setIsLoading(true);
+    setError(null);
+
+    try {
```
DanielMoraDC, line 33: I would suggest to restrict the try/catch to the `playlistService` call. By doing so, it is more explicit to understand which parts of the code are expected to fail.

barroro: It's fine but we should change the approach in the rest of the hooks, because I've implemented using try/catch in the hook instead of service.

src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -0,0 +1,79 @@
...
+        };
+      } else {
+        // Assuming playlistCreationMode is PlaylistCreationMode.Existing
+        return {
+          success: true,
+          data: 'mock-playlist-id',
+        };
+      }
+    } catch (err) {
+      setError(
```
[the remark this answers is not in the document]
barroro: True, I'm going to change the first point. What do you mean in the second point? I can remove the `setError` because right now it doesn't use it.

src/pages/api/playlists/create.ts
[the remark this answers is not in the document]
barroro: Done!

src/components/generate/usePlaylistSubmission.tsx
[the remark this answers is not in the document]
barroro: Done the first point! 👌

src/components/generate/PlaylistGetUrlLink/PlaylistGetUrlLink.tsx
```diff
@@ -6,11 +6,16 @@ import playlistReadyImage from '@public/playlist-ready.svg';
...
 
-const PlaylistGetUrlLink = () => {
+const PlaylistGetUrlLink = ({
+  playlistId,
+}: {
+  playlistId: string | undefined;
+}) => {
   const { t } = useTranslation('generate');
   const { copy, isCopied } = useCopyToClipboard();
+  const urlWithId = `${URL_BASE}${playlistId}`;
```
DanielMoraDC, line 18: We can do this in a separate PR, but taking the chance to share what we discussed earlier. We can simply embed a Playlist using something like this:

```ts
      <iframe 
        style={{ borderRadius: '12px' }} 
        src={`https://open.spotify.com/embed/playlist/${playlistId}?utm_source=generator`}
        width="100%" 
        height="352" 
        allowFullScreen 
        allow="autoplay; clipboard-write; encrypted-media; fullscreen; picture-in-picture" 
        loading="lazy"
      ></iframe>
```

From the [Spotify docs](https://developer.spotify.com/documentation/embeds). This is the simplest way to embed it, but there are more sophisticated ones in the link. Here is how it looks like out of the box:


https://github.com/user-attachments/assets/f3022e02-4935-4916-aeb7-f0c73f597a36

src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -61,35 +65,46 @@ export type FormSchemaType = z.infer<typeof formSchema>;
...
-  const onSubmit = (_values: FormSchemaType) => {
-    setCurrentStep((prev) => prev + 1);
+  const onSubmit = async (values: FormSchemaType) => {
+    const { success, data: playlistId } = await submitPlaylist(values);
+
+    if (success) {
+      setCurrentStep((prev) => prev + 1);
+      setPlaylistId(playlistId);
+    } else {
+      toast.error(t('steps.errors.createNewPlaylist.unexpectedError'));
```
DanielMoraDC, line 106: Probably a nitpick, but I believe it would be cleaner to move the actual trigger of the toast inside the return, so this function only changes the internal state without interacting with UI elements.

locales/en/generate.json
```diff
@@ -1,70 +1,77 @@
...
+      "description": "Create a new list or use an existing one",
+      "form": {
+        "createNewPlaylist": {
+          "description": "Use this option to create a new Spotify playlist",
+          "descriptionPlaceholder": "Playlist description",
+          "giveADescription": "Give a description to the playlist",
+          "giveAName": "Give a name to the playlist",
+          "namePlaceholder": "Playlist name",
+          "publicPlaylist": {
+            "description": "The playlist will be public and you will be able to find it in your library",
```
DanielMoraDC, line 40: I think this can be confusing, as both private and public will show in the library. However, only public ones will appear in the profile. I would suggest something similar to:

```
Private playlists won't be visible in your profile.
```

__tests__/generate.test.tsx
```diff
@@ -286,10 +308,122 @@ describe('GeneratePlaylistPage', () => {
...
+      artists: [
+        {
+          name: 'Holding Absence',
+        },
+      ],
+    });
+  });
+
+  it('should copy the URL link into the clipboard when clicking the copy button', async () => {
+    const spyClipboardWriteText = vi
```
DanielMoraDC, line 326: This test is a tiny variation of the first one, which is already significantly large. Is there a way to either reuse the arrange blocks or simply let the test force to start from a specific step?

barroro: We can create blocks to make the tests more readable, or we can move this logic into the first one. I started putting the copy clipboard check in the same test, but I tried to separate it into different scenarios.

DanielMoraDC: Feel free to also move this to a separate task to implement after merging.

Self-resolve at will.

src/components/generate/usePlaylistSubmission.tsx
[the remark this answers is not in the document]
DanielMoraDC: About the second point: I thought we were forwarding the error somehow into the UI, so I thought we might have to localize them. But I see now that the error message is actually not used (should be just remove it?).

We are just returning true/false depending on whether an error is raised, so I think it should be better to just let the code faill and let the caller handle the error. Not only it adds extra complexity but it also hides the original error traces.

src/components/generate/usePlaylistSubmission.tsx
DanielMoraDC: Resolving this, as it overlaps with https://github.com/festwrap/festwrap-ui/pull/33#discussion_r2048376061

__tests__/generate.test.tsx
barroro: Perfect! I will create a task to review this part 👌

DanielMoraDC: Created task, resolving.

locales/en/generate.json
barroro: Done! I've modified the description.

locales/es/generate.json
```diff
@@ -37,7 +37,7 @@
           "giveAName": "Da un nombre a la lista de reproducción",
           "namePlaceholder": "Nombre de la lista de reproducción",
           "publicPlaylist": {
-            "description": "La lista de reproducción será pública y podrás encontrarla en tu biblioteca",
+            "description": "Solo las listas de reproducción pública son visibles en su perfil",
```
DanielMoraDC, line 40: Nitpick: I would not be so formal 😅  "su" -> "tu"

locales/ca/generate.json
```diff
@@ -37,7 +37,7 @@
           "giveAName": "Dona un nom a la llista",
           "namePlaceholder": "Nom de la llista",
           "publicPlaylist": {
-            "description": "La llista de reproducció serà pública i la podràs trobar a la teva biblioteca",
+            "description": "Només les llistes de reproducció públiques són visibles al vostre perfil",
```
DanielMoraDC, line 40: Same as in https://github.com/festwrap/festwrap-ui/pull/33/commits/f0b7c84094801ce81e9fed91dbae7d34a67053dd#r2050486372

barroro: Done!

src/components/generate/usePlaylistSubmission.tsx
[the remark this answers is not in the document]
DanielMoraDC: As agreed, we will resolve for now discuss separately.

DanielMoraDC (approved):

PR #35, opened 2025-04-18
feat: FW-100 Add list in success step

DanielMoraDC:
# Description

Replaces url copy step with list visualization.

Next steps:
- Create error page when playlist update fails.
- Add error messages on last step when partial errors occur (e.g. some artists could not be added).

# Testing

https://github.com/user-attachments/assets/f96bb8ac-4aea-4186-b487-08de851e58c5

### DIFF
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -125,6 +125,15 @@ const customRenderWithProviders = (ui: ReactNode) => {
 };
 
 describe('GeneratePlaylistPage', () => {
+
+  let searchResultArtists = ['Holding Absence'];
+
+  beforeEach(() => {
+    artistsService.searchArtists.mockResolvedValue({
+      artists: searchResultArtists.map(name => ({ name: name, imageUri: null }))
+    });
+  });
+
   it('should render the form with the first step displayed', async () => {
     customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
 
@@ -251,14 +260,6 @@ describe('GeneratePlaylistPage', () => {
   });
 
   it('should create the new playlist and display the success message', async () => {
-    artistsService.searchArtists.mockResolvedValue({
-      artists: [
-        {
-          name: 'Holding Absence',
-          imageUri: null,
-        },
-      ],
-    });
     playlistsService.createNewPlaylist.mockResolvedValue({
       playlistCreated: {
         id: '123',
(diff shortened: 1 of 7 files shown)


src/components/generate/PlaylistUpdateReport/PlaylistOverview.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const PlaylistOverview: FC<PlaylistOverviewProps> = ({
+  playlistId,
+  height = 500,
+}) => {
+  const embedUrl = `https://open.spotify.com/embed/playlist/${playlistId}`;
+
+  return (
+    <iframe
+      src={embedUrl}
+      style={{ borderRadius: '12px' }}
```
barroro, line 17: We use TailwindCSS for the styles, in this case, you have to replace for `className="rounded"` instead. Check the documentation: [Link](https://tailwindcss.com/docs/border-radius)

DanielMoraDC: Done ✅

__tests__/generate.test.tsx
```diff
@@ -368,32 +350,13 @@ describe('GeneratePlaylistPage', () => {
       const thirdStepContent = screen.getByRole('tabpanel');
       return within(thirdStepContent).getByText(/steps.step3.title/i);
     });
+    const embeddedPlaylist = screen.queryByTitle('Spotify embedded playlist');
```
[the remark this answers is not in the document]
DanielMoraDC: `queryByTitle` returns `None` if not found while `findByTitle` raises an error if not found. Therefore, I am using `queryByTitle` when we were checking expectations. Using `findByTitle` already does both but it is less explicit in my point of view.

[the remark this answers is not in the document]
barroro: Cool! I wanted to know your opinion. I usually use `findByTitle` when the element must be on the screen and `queryByTitle`, when I have to check an element that doesn't exist on the screen. I was researching, and the difference between them is when you expect an async element or not as well. But I agree with this solution 👌

__tests__/generate.test.tsx
```diff
@@ -125,6 +125,20 @@ const customRenderWithProviders = (ui: ReactNode) => {
 };
 
 describe('GeneratePlaylistPage', () => {
+
+  let searchResultArtists = [{ name: 'Holding Absence', imageUri: null }];
+  let createdPlaylistId = '123';
+
+  beforeEach(() => {
+    artistsService.searchArtists.mockResolvedValue({
+      artists: searchResultArtists.map(artist => ({ name: artist.name, imageUri: artist.imageUri })),
```
barroro, line 134: This mapper is unnecessary right now, you are mapping the same value. We could return just the value of `searchResultArtists`. What do you think?

DanielMoraDC: I made the change without thinking, you are absolutely right. Done ✅

barroro (approved):
Good job!

PR #37, opened 2025-04-24
feat: [FW-92] Create new API endpoint to update playlist

barroro:
## Description
Define a new flow in the API to update playlists, adding new songs to the playlist based on the selected artists.

## Changes
- Add a new method in the playlist client called `updatePlaylist`.
- Add new API route `api/playlists/update/[ID]`
- Add test cases

## Testing
- You have to take the cookie from the request using the client app and executing for any GET endpoint from the request in Chrome Devtools.

image

```
curl --location 'http://localhost:3000/api/playlists/update/[ID]' \
--header 'Cookie: <NEXT-AUTH-TOKEN>' \
--header 'Content-Type: application/json' \
--data '{
    "artists": [
        { "name": "Holding Absence" },
        { "name": "Attila" },
        { "name": "Polaris" },
        { "name": "Parkway Drive" }
    ]
}'
```


src/lib/clients/playlists.test.ts
```diff
@@ -259,4 +260,131 @@ describe('PlaylistsHTTPClient', () => {
...
+        playlistId: '123',
+        artists: [{ name: 'Artist 3' }, { name: 'Artist 4' }],
+      };
+      authHeaderBuilder = new AuthHeaderBuilderStub();
+    });
+
+    const mockHttpClientResponse = (status: number = 200) => {
+      response = {
+        data: {
+          playlist: {
```
DanielMoraDC, line 285: To make this consistent with the actual implementation, we should return only the status.

src/lib/clients/playlists.test.ts
```diff
@@ -259,4 +260,131 @@ describe('PlaylistsHTTPClient', () => {
       ).rejects.toThrow();
     });
   });
+
+  describe('updatePlaylist', () => {
```
DanielMoraDC, line 264: I feel we are duplicating most of the test code, here. Is there a way we can reuse most of this code?

src/pages/api/playlists/update/[playlistId].ts
```diff
@@ -0,0 +1,106 @@
...
+const updatePlaylistSchema = z.object({
+  playlistId: z.string(),
+  artists: z.array(
+    z.object({
+      name: z.string(),
+    })
+  ),
+});
+
+export function createUpdatePlaylistHandler({
```
Comment (DanielMoraDC) [inline src/pages/api/playlists/update/[playlistId].ts:30]: This handler is also mostly a duplicate of the existing one. Do you think we can reuse some code without adding too much complexity?

DanielMoraDC:
I believe using `api/playlists/{id}` is a more Restful way to name the endpoint. The `PUT` method already implies an update. Moreover, by doing it like this, we keep consistency with the backend.

src/lib/clients/playlists.test.ts
barroro: I've added some changes to reuse part of the code in the three scenarios. 👌

DanielMoraDC: I find the new approach makes it difficult to understand what is being tested.

I came up with a proposal, which basically reuses the arrange stage of each test. I moved the common elements into a shared test setup and then let each tests extend that with their own specific variables. Here is the commit, lEt me know what you think: https://github.com/festwrap/festwrap-ui/commit/b2e908c8b33fec0eaa88abddc383d802a87a6e4e

src/pages/api/playlists/update/[playlistId].ts
Comment (DanielMoraDC) [reply src/pages/api/playlists/update/[playlistId].ts:30]: I am fine duplicating this since the amount of code is small. I created a task in the backlog to review this back later.

src/pages/api/playlists/update/update.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    await createHandler({ client })(request, createMockResponse());
+
+    expect(client.updatePlaylist).toHaveBeenCalledWith('mocked-token', {
+      playlistId: playlistData.playlistId,
+      artists: playlistData.artists,
+    });
+  });
+
+  it('should handle query parameters when provided', async () => {
```
DanielMoraDC, line 102: I am not sure what this tests is checking, as the assert section does not match this description. Should we remove it?

barroro: Totally agree. It doesn't make sense this test. Done! 👌

src/lib/handlers/createBaseHandler.ts
```diff
@@ -0,0 +1,60 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
```
DanielMoraDC, line 1: Probably a nitpick, but I would rather name this file something like: `src/lib/handlers/base.ts`, as `handler` is already implicit in the folder name and  `createBaseHandler` is the name of the function.

src/lib/handlers/createBaseHandler.ts
```diff
@@ -0,0 +1,60 @@
...
+  validationSchema: ZodSchema;
+  handleRequest: (
+    _requestData: TRequestData,
+    _accessToken: string,
+    _response: NextApiResponse<TResponseData>
+  ) => Promise<void>;
+  extractRequestData: (_req: NextApiRequest) => any;
+};
+
+export function createBaseHandler<
```
DanielMoraDC, line 23: If we use this as base handler for all handlers, then we should also migrate the existing search playlist endpoint to this. If this is only meant to reuse code for the update/create playlist, then I would reflect that in the name and I would also move the missing artist condition in here.

src/lib/handlers/createBaseHandler.ts
barroro: Done!

src/lib/handlers/createBaseHandler.ts
barroro: I totally agree! I'm going to apply to the rest of the endpoints 👌

barroro: Done!

src/pages/api/artists/search.ts
```diff
@@ -42,32 +43,16 @@ export function createSearchArtistHandler({
...
+  const searchQuerySchema = createSearchQuerySchema(defaultLimit, maxLimit);
 
-    if (!token) {
-      response.status(401).json({ message: 'Unauthorized' });
-      return;
-    }
+  return createBaseHandler<typeof searchQuerySchema._type, ResponseData>({
+    validationSchema: searchQuerySchema,
+    extractRequestData: (req) => req.query,
+    handleRequest: async (requestData, accessToken, response) => {
```
DanielMoraDC, line 51: I feel this part is a bit difficult to read. Could we declare the handler function in its own function?

src/pages/api/playlists/create.ts
```diff
@@ -34,37 +34,18 @@ const createNewPlaylistSchema = z.object({
...
-  ): Promise<void> {
-    const body = request.body;
-    const parsedArgs = createNewPlaylistSchema.safeParse(body);
+  return createBaseHandler<
+    typeof createNewPlaylistSchema._type,
+    CreatePlaylistResponseData
+  >({
+    validationSchema: createNewPlaylistSchema,
+    extractRequestData: (req) => req.body,
+    handleRequest: async (requestData, accessToken, response) => {
```
DanielMoraDC, line 43: Same for this one.

src/pages/api/playlists/search.ts
```diff
@@ -42,32 +43,16 @@ export function createSearchPlaylistHandler({
...
+  const searchQuerySchema = createSearchQuerySchema(defaultLimit, maxLimit);
 
-    if (!token) {
-      response.status(401).json({ message: 'Unauthorized' });
-      return;
-    }
+  return createBaseHandler<typeof searchQuerySchema._type, ResponseData>({
+    validationSchema: searchQuerySchema,
+    extractRequestData: (req) => req.query,
+    handleRequest: async (requestData, accessToken, response) => {
```
DanielMoraDC, line 51: And also here 😅

src/pages/api/playlists/update/[playlistId].ts
```diff
@@ -0,0 +1,78 @@
...
+  client,
+}: UpdatePlaylistHandlerParams) {
+  return createBaseHandler<
+    typeof updatePlaylistSchema._type,
+    UpdatePlaylistResponseData
+  >({
+    validationSchema: updatePlaylistSchema,
+    extractRequestData: (req) => ({ ...req.body, ...req.query }),
+    handleRequest: async (requestData, accessToken, response) => {
```
Comment (DanielMoraDC) [inline src/pages/api/playlists/update/[playlistId].ts:39]: And here.

src/pages/api/artists/search.ts
barroro: What do you mean? Do you refer to moving the handler before calling `createBaseHandler`?

DanielMoraDC: Sorry I was not very specific. My point is that there are three different nesting levels, which makes it difficult to read. I am proposing something like this (created with Copilot):

```ts
async function handleSearchArtists(
  requestData: { name: string; limit: number },
  accessToken: string,
  client: ArtistsClient
) {
  const { name, limit } = requestData;
  const searchResults = await client.searchArtists(accessToken, name, limit);
  
  return {
    artists: searchResults.map((artist: Artist) => ({
      name: artist.getName(),
      imageUri: artist.getImageUri(),
    })),
    message: 'Artists successfully retrieved',
  };
}

export function createSearchArtistHandler({
  client,
  defaultLimit,
  maxLimit,
}: SearchArtistHandlerParams) {
  const searchQuerySchema = createSearchQuerySchema(defaultLimit, maxLimit);
  
  return createBaseHandler<typeof searchQuerySchema._type, ResponseData>({
    validationSchema: searchQuerySchema,
    extractRequestData: (req) => req.query,
    handleRequest: async (requestData, accessToken, response) => {
      const result = await handleSearchArtists(requestData, accessToken, client);
      response.status(200).json(result);
    },
  });
}
```

barroro: Ok, in this approach could have make sense but what happen in the `create` scenario? In that scenario we are returning different responses depending on the status received. Would we have separate logic between the handler function and the responses? 🤔

DanielMoraDC: Created a task for this, resolving for now.

DanielMoraDC (approved):

PR #38, opened 2025-05-01
refactor(tests): optimize GeneratePlaylistPage test suite for performance

barroro:
## Description
Improve performance and readability in `generate` test file.

## Changes
- Reduce async operations and wait times by consolidating helper functions
- Create centralized action helpers and test data objects
- Add proper TypeScript interfaces for service mocks
- Replace redundant beforeEach with more efficient setup/teardown
- Optimize user interactions by removing artificial delays
- Improve test organization with focused describe blocks
- Enhance maintainability with better variable naming and documentation


__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-    expect(itemOption).toBeInTheDocument();
-    userEvent.click(itemOption);
-  });
-};
+interface PlaylistsService {
+  searchPlaylists: ReturnType<typeof vi.fn>;
+  createNewPlaylist: ReturnType<typeof vi.fn>;
+}
+
+interface ArtistsService {
```
DanielMoraDC, line 78: This interface already exists [here](https://github.com/festwrap/festwrap-ui/blob/main/src/services/artistsService.ts#L8). Why redifining it here?

DanielMoraDC: Same for `PlaylistsService`.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
+  searchArtists: ReturnType<typeof vi.fn>;
+}
+
+interface ServiceMocks {
```
DanielMoraDC, line 82: For consistency, I would use a `type` instead of an interface as it only encapsulates some data local to this file.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-  await waitFor(() => {
-    const itemOption = screen.getByRole('option', {
-      name: artistName,
-    });
-    expect(itemOption).toBeInTheDocument();
-    userEvent.click(itemOption);
-  });
+const TEST_DATA = {
+  playlists: {
+    default: {
```
DanielMoraDC, line 52: Nitpick: instead of default, we can use `existing` to make it consistent.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
+      description: 'Existing playlist description',
+      isPublic: true,
+    },
+    new: {
+      name: 'Fest Bops',
+      description: 'Generated by FestWrap',
+    },
+  },
+  artists: {
+    default: { name: 'Holding Absence', imageUri: null },
```
DanielMoraDC, line 64: Same here. Using `single` would be provide more information.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
+afterAll(() => {
+  vi.restoreAllMocks();
+});
 
-const MockServiceProvider = ({ children }: { children: ReactNode }) => {
-  return <ServiceProvider value={mockServices}>{children}</ServiceProvider>;
-};
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
```
DanielMoraDC, line 116: It does not feel consistent that the mock creation happens inside the `describe` but the rest are declared in the parent scope. Shouldn't we use them in the same scope?

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
+beforeAll(() => {
+  user = userEvent.setup({ delay: null });
+});
 
-const mockServices = {
-  playlistsService,
-  artistsService,
-};
+afterAll(() => {
+  vi.restoreAllMocks();
```
DanielMoraDC, line 111: Instead of manually doing this, I believe we can make mocks be restored on each implementation defining that in the config. See this: https://vitest.dev/config/#restoremocks

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-    await clickToNextButton();
-
-    const playlistNameError = await waitFor(() => {
-      return screen.getByText(/steps.errors.name.required/i);
-    });
-
-    expect(playlistNameError).toBeInTheDocument();
-  });
+const actions = {
+  async click(selector: RegExp | string) {
```
DanielMoraDC, line 127: I see the return type is never used, so I would rather not return anything.

DanielMoraDC: Actually, none of the code blocks using these actions use the return value, so we can remove that from all.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
 
-    await clickToNextButton();
-
-    const secondStepContentTitle = await waitFor(() => {
-      const secondStepContent = screen.getByRole('tabpanel');
-      return within(secondStepContent).getByText(/steps.step2.title/i);
+    await waitFor(async () => {
+      const option = screen.getByRole('option', { name: artistName });
+      await user.click(option);
+      expect(screen.getByText(artistName)).toBeInTheDocument();
```
DanielMoraDC, line 196: I think it is better that these methods do not perform any expectations and only prepare data. By moving expectations here, it gets more difficult to understand what each test is checking.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
   });
 
-  it('should create the new playlist and display the success message', async () => {
-    customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
-
-    const playlistNameInput = screen.getByLabelText(
-      /steps.step1.form.createNewPlaylist.giveAName/i
-    );
-    await userEvent.type(playlistNameInput, 'My new playlist');
+  it('should navigate between steps correctly', async () => {
```
DanielMoraDC, line 229: A couple of comments on this test:
- It is not clear what it is the arrange, the act and the assert here 🤔 
- It is only doing navigation from first to second step. Should  we reflect that in the test name?

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-    const thirdStepContentTitle = await waitFor(() => {
-      const thirdStepContent = screen.getByRole('tabpanel');
-      return within(thirdStepContent).getByText(/steps.step3.title/i);
-    });
+      await actions.completeFirstStepNewPlaylist(name, description);
+      await actions.completeSecondStep(artistsToSelect);
 
-    expect(thirdStepContentTitle).toBeInTheDocument();
+      expect(
+        screen.getByText(/steps.step3.playlisyGeneratedSuccessfully/i)
```
DanielMoraDC, line 249: Querying by text feel a bit brittle. I found [this recommend order of prioritization](https://testing-library.com/docs/queries/about/#priority) for querying DOM elements. Roles and Labels are placed first than text. What are your thoughts?

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-    expect(playlistsService.createNewPlaylist).toHaveBeenCalledWith({
-      playlist: {
-        name: 'My new playlist',
-        description: 'My new playlist description',
-        isPublic: false,
-      },
-      artists: [
-        {
-          name: 'Holding Absence',
+      expect(
```
DanielMoraDC, line 257: Thinking out loud: do we need to check this? Given we are focusing here in the rendered components, not sure if this assertion is very useful in this context.

DanielMoraDC: I am actually not sure about this, as it might be useful to know that we are properly creating it. Wondering if we might also want to parametrize the public/private toggle for greater coverage, then.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-
-    const secondStepContentTitle = await waitFor(() => {
-      const secondStepContent = screen.getByRole('tabpanel');
-      return within(secondStepContent).getByText(/steps.step2.title/i);
+        artists: artistsToSelect.map((name) => ({ name })),
+      });
     });
 
-    expect(secondStepContentTitle).toBeInTheDocument();
+    it('should handle multiple artist selection and removal', async () => {
```
DanielMoraDC, line 269: This is not exclusive to the new playlist flow, so maybe we can move the artist-related checks to the root context.

__tests__/generate.test.tsx
```diff
@@ -38,453 +47,308 @@ const staticTranslations: GenerateProps = {
...
-      ],
-    });
-
-    customRenderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
-
-    await completeFirstStepNewPlaylist();
-    expect(await waitForSecondStepContent()).toBeInTheDocument();
+  describe('Existing Playlist Flow', () => {
+    it('should successfully update an existing playlist', async () => {
+      mockServices.playlistsService.createNewPlaylist.mockResolvedValue({
```
DanielMoraDC, line 324: Is this necessary? Does not seem to relate to the test.

__tests__/generate.test.tsx
barroro: I agree! I'm going to unify the part where are using interfaces 👌

__tests__/generate.test.tsx
barroro: I'm agree with you, but there are some cases where you need to add the expectation to cover all the step/flow. For example, in this case for me it makes sense, because it's the way to ensure that the option has been selected correctly.
The most important part in each test is the expectation on the bottom of each test, it's the goal of each test or at least that's how I see it.

__tests__/generate.test.tsx
barroro: Do you talk about this test? I think that it's very simple this test and I don't understand where is the problem 🤔 
```
it('should navigate between steps correctly', async () => {

    await actions.completeFirstStepNewPlaylist();

    await actions.click(/steps.navigation.previous/i);
    await actions.waitForStep(/steps.step1.title/i);
  });
```

__tests__/generate.test.tsx
barroro: What query are you referring to? Are you referring to the successful message?

__tests__/generate.test.tsx
barroro: For me it makes sense to check the context with you are calling the create method, but it couldn't make sense to check it if you already checked the UI.

__tests__/generate.test.tsx
barroro: Sure, but I would change when the code be duplicated in the next PR. If we thinking in the current need, we have just one option available right?

__tests__/generate.test.tsx
barroro: I think that it's a mistake. Let me check and I'm going to fix it!

__tests__/generate.test.tsx
barroro: Cool! Let me check and I will try to implement 👌

__tests__/generate.test.tsx
barroro: Sure, we can move inside the describe for more consistency 👌

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: Done in all actions! 👌

__tests__/generate.test.tsx
barroro: I've renamed the test to be more descriptive 👌

__tests__/generate.test.tsx
barroro: Done! Good point. I normally avoid use `getByText` but in this case, as it was a single text, I did not take it into account. I have changed the approach by adding the text as a title to be able to use the `getByRole` with `heading`.

__tests__/generate.test.tsx
DanielMoraDC: I do not see any changes. Did you push? 🤔

__tests__/generate.test.tsx
DanielMoraDC: I totally agree that tests should end up with the expectations, but I do not feel the same about auxiliar methods. If this expectation fails, then all tests that use this helper will fail, which feels wrong 🤔 

It also helps understand what the test checks by keeping all expects in the test. An alternative is also change the name of the method so it is clear that it does not only the arrange but also checks for expectations (e.g. selectArtistAndAssertSelected).

__tests__/generate.test.tsx
DanielMoraDC: The name feels much clearer now :)

__tests__/generate.test.tsx
barroro: Sorry, I didn't understand the comment correctly and I changed the interface to a type. I'm going to remove it and use the real interfaces 👌

__tests__/generate.test.tsx
DanielMoraDC: About the first point, it is not clear what the test is actually checking:

```ts
  it('should navigate to the first step clicking on the previous button', async () => {

    await actions.completeFirstStepNewPlaylist();

    await actions.click(/steps.navigation.previous/i);
    await actions.waitForStep(/steps.step1.title/i);
  });
```

There are no explicit references to `expect` and the `waitForStep` action does not indicate that is doing any checks unless you actual check the code.

Resolve at will.

__tests__/generate.test.tsx
DanielMoraDC: Great, resolving!

__tests__/generate.test.tsx
DanielMoraDC: It's ok to do it in the next one, let's make sure to do it 🕵️

__tests__/generate.test.tsx
DanielMoraDC: We can iterate this in the next PRs. Resolving

__tests__/generate.test.tsx
barroro: Done! I've changed the approach. Now we are using the real interfaces, and when you have to use in any test, you have to wrap the method with `vi.mocked` instead

__tests__/generate.test.tsx
barroro: I understand you. It's a bit strange but if you thinking in how `waitFor` method works you are waiting for something. So, in this case, it could make sense this approach. We could return the element in `waitForStep` and do the expectation from the test. What do you think?

__tests__/generate.test.tsx
barroro: I see more of the second option. We could change the name to be more semantic 👌

barroro: Done! I've updated the `selectArtist` and `selectPlaylist` methods. I have some doubts about the complete steps because the name "complete" already describes the action of moving to the next step, right? What do you think?

__tests__/generate.test.tsx
DanielMoraDC: That makes sense!

barroro: Done! I've changed the test to do the expectation inside the test even though the action function is already checking it.

DanielMoraDC: I truly believe we should move the expects outside the arrangement functions, it makes tests more difficult to debug.

I am resolving for now, I created a task to handle this so we can unblock this PR.

DanielMoraDC (approved):

PR #39, opened 2025-05-06
chore: remove copy token to clipboard

DanielMoraDC:
# Description

Removes the _copy token to clipboard_ functionality, as it was a development feature that should not be included in production.


barroro (approved):
Good job!

PR #40, opened 2025-05-07
feat: [FW-109] submit generate form to add songs in a existing playlist

barroro:
### Description
Allow submission of the form using the existing playlist option.


src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -24,11 +25,12 @@ export function usePlaylistSubmission(): UsePlaylistSubmissionResult {
...
   ): Promise<SubmitPlaylistResponse> => {
     const { playlistCreationMode, ...playlistData } = values;
+    let response: SubmitPlaylistResponse;
 
     setIsLoading(true);
 
-    try {
-      if (playlistCreationMode === PlaylistCreationMode.New) {
+    if (playlistCreationMode === PlaylistCreationMode.New) {
+      try {
```
DanielMoraDC, line 33: The only thing that should be raising an error is the playlist creation/update, so moving the try to that specific scope is more convenient and will make code easier to read and maintain.

src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -40,32 +42,51 @@ export function usePlaylistSubmission(): UsePlaylistSubmissionResult {
...
           success: true,
-          data: 'mock-playlist-id',
+          data: playlistData.playlistSelected?.id,
         };
-      } else {
-        return {
+      } catch (error) {
+        response = {
           success: false,
+          errorKey: 'steps.errors.existingPlaylist.unexpectedError',
```
DanielMoraDC, line 78: The status does not seem to be properly encapsulated, as we have an implicit dependency between `success` and `errorKey` (errorKey should be null when success is true). What about returning a `status` enum, instead?

src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -40,32 +42,51 @@ export function usePlaylistSubmission(): UsePlaylistSubmissionResult {
...
+      } catch (error) {
+        response = {
+          success: false,
+          errorKey: 'steps.errors.createNewPlaylist.unexpectedError',
+        };
+      }
+    } else if (playlistCreationMode === PlaylistCreationMode.Existing) {
+      try {
+        const existingPlaylistData: UpdatePlaylistDTO = {
+          playlistId: playlistData.playlistSelected?.id || '',
```
DanielMoraDC, line 63: Why passing empty, here?

src/components/generate/usePlaylistSubmission.tsx
```diff
@@ -9,6 +9,7 @@ import {
 type SubmitPlaylistResponse = {
   success: boolean;
   data?: string | undefined;
```
DanielMoraDC, line 11: It can also be a good chance to change this attribute name, as It is the `playlistId`, so we are more explicit.

__tests__/generate.test.tsx
```diff
@@ -323,5 +324,25 @@ describe('GeneratePlaylistPage', () => {
         await screen.findByText(/steps.errors.playlistSelected.required/i)
       ).toBeInTheDocument();
     });
+
+    it('should display error toast when API fails', async () => {
```
DanielMoraDC, line 328: We have a nearly identical test for new playlists, shouldn't we reuse this?. Moreover, we have some tests in the new playlist section that also apply to the existing playlist.

I would suggest we have tests that applies to both cases (parametrized so we cover both pasts) and then keep the current two sections for tests specific to each flow.

src/components/generate/usePlaylistSubmission.tsx
barroro: It should never be empty. I know it's strange, but it's for the form schema implementation. In the schema is an optional field depending on the initial form selection. I'm going to check if there's a better approach to avoid this condition.

src/components/generate/usePlaylistSubmission.tsx
barroro: I'm agree. I'm going to change it 👌

__tests__/generate.test.tsx
barroro: I'm not sure that the tests are nearly identical. The implementations are quite different, and it's more readable to have a specific test for each scenario. We could create a new `action` using the rest of the actions but also the fields that you have to fill are different in some cases. I don't see a good way to resolve that. What solution were you thinking?

src/components/generate/usePlaylistSubmission.tsx
barroro: Ok, I was thinking to show a more friendly and specific for each flow, but we could use the generic one and avoid create multiple `try/catch`.

src/components/generate/usePlaylistSubmission.tsx
barroro: Related to the previous conversation [link](https://github.com/festwrap/festwrap-ui/pull/40#discussion_r2078350059), it wouldn't make sense to return `errorKey` if we use the same error for both flows, right?

src/components/generate/usePlaylistSubmission.tsx
barroro: Done!

src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -18,47 +18,35 @@ const STEPS_COUNT = 3;
...
-        code: z.ZodIssueCode.custom,
-        message: 'Playlist is required for EXISTING role',
-        path: ['playlistSelected'],
-      });
-    }
-  });
+} as const;
+
+const baseSchema = z.object({
+  description: z.string().optional(),
```
DanielMoraDC, line 24: Shouldn't description and isPublic be in the new one, only? The existing playlist should only contain the id, right?

src/components/generate/usePlaylistSubmission.tsx
DanielMoraDC: This is what I am proposing:

```ts
import { useState } from 'react';
import { CreatedPlaylistStatus, CreateNewPlaylistDTO, UpdatePlaylistDTO } from '@/entities/playlists';
import { useServices } from '@/contexts/ServiceContext';
import {
  FormSchemaType,
  PlaylistCreationMode,
} from '@/components/generate/GeneratePlaylistStepper';

export enum SubmissionStatus {
  OK = 'OK',
  PARTIAL_ERRORS = 'SOME_ERRORS',
  ERROR = 'ERROR',
}

type SubmitPlaylistResponse = {
  status: SubmissionStatus;
  playlistId?: string | undefined;
  errorKey?: string;
};

function dtoStatusToSubmissionStatus(
  status?: CreatedPlaylistStatus
): SubmissionStatus {
  switch (status) {
    case CreatedPlaylistStatus.OK:
      return SubmissionStatus.OK;
    case CreatedPlaylistStatus.MISSING_ARTISTS:
      return SubmissionStatus.PARTIAL_ERRORS;
    default:
      return SubmissionStatus.ERROR;
  }
}

interface UsePlaylistSubmissionResult {
  isLoading: boolean;
  submitPlaylist: (_values: FormSchemaType) => Promise<SubmitPlaylistResponse>;
}

export function usePlaylistSubmission(): UsePlaylistSubmissionResult {
  const { playlistsService } = useServices();
  const [isLoading, setIsLoading] = useState(false);

  const submitPlaylist = async (
    values: FormSchemaType
  ): Promise<SubmitPlaylistResponse> => {
    let response: SubmitPlaylistResponse = {
      status: SubmissionStatus.ERROR
    };

    setIsLoading(true);

    if (values.playlistCreationMode === PlaylistCreationMode.New) {
      const newPlaylistData: CreateNewPlaylistDTO = {
        playlist: {
          name: values.name,
          description: values.description,
          isPublic: values.isPublic,
        },
        artists: values.artists.map((artist) => ({
          name: artist,
        })),
      };

      try {
        const serviceResponse = await playlistsService.createNewPlaylist(newPlaylistData);
        response = {
          // I believe the status should be moved to the top level to be consistent with the existing playlist creation
          status: dtoStatusToSubmissionStatus(serviceResponse.playlistCreated?.status),
          playlistId: serviceResponse.playlistCreated?.id,
        };
        } catch (error) {
        response = {
          status: SubmissionStatus.ERROR,
        };
      }
    } else if (values.playlistCreationMode === PlaylistCreationMode.Existing) {
      const playlistId = values.playlistSelected.id
      const existingPlaylistData: UpdatePlaylistDTO = {
        playlistId: playlistId,
        artists: values.artists.map((artist) => ({
          name: artist,
        })),
      };

      try {
        const serviceResponse = await playlistsService.updatePlaylist(existingPlaylistData);
        response = {
          status: dtoStatusToSubmissionStatus(serviceResponse.status),
          playlistId: playlistId,
        };
        } catch (error) {
          response = {
            status: SubmissionStatus.ERROR,
          };
      }
    }

    setIsLoading(false);
    return response;
  };

  return {
    isLoading,
    submitPlaylist,
  };
}
```

I see a couple of problems:
- Probably too much overhead to translate the DTO status to another status, but I do not think we should be passing the DTO object all around.
- Status should be consistently placed in the same level for existing and new playlists. For new playlists, it is inside `playlistCreated` while for existing it is in the top level.

I did not include error messages here so it is the UX component and not the submission hook to decides what to display.

What do you think?

src/components/generate/usePlaylistSubmission.tsx
DanielMoraDC: See proposal here: https://github.com/festwrap/festwrap-ui/pull/40#discussion_r2079057350

__tests__/generate.test.tsx
DanielMoraDC: This works, for example:

```ts
    it.each([
      {
        scenario: 'existing playlist update',
        getUpdatePlaylistFn: () => mockServices.playlistsService.updatePlaylist,
        firstStepFn: actions.completeFirstStepExistingPlaylist,
        errorMessage: 'steps.errors.existingPlaylist.unexpectedError'
      },
      {
        scenario: 'new playlist creation',
        getUpdatePlaylistFn: () => mockServices.playlistsService.createNewPlaylist,
        firstStepFn: actions.completeFirstStepNewPlaylist,
        errorMessage: 'steps.errors.createNewPlaylist.unexpectedError'
      }
    ])('should display error toast when $scenario API fails', async ({ getUpdatePlaylistFn, firstStepFn, errorMessage }) => {
      vi.mocked(getUpdatePlaylistFn()).mockRejectedValue(
        new Error('API Error')
      );

      await firstStepFn();
      await actions.selectArtistAndAssertSelected(TEST_DATA.artists.single.name);
      await actions.click(/steps.navigation.generate/i);

      await waitFor(() => {
        expect(toast.error).toHaveBeenCalledWith(errorMessage);
      });
    });
  });
```

DanielMoraDC: Same for this other one:

```ts
    it.each([
      {
        scenario: 'existing playlist update',
        firstStepFn: actions.completeFirstStepExistingPlaylist,
      },
      {
        scenario: 'new playlist creation',
        firstStepFn: actions.completeFirstStepNewPlaylist,
      }
    ])('should handle multiple artist selection and removal for $scenario', async ({ firstStepFn }) => {
      vi.mocked(mockServices.artistsService.searchArtists).mockResolvedValue({
        artists: TEST_DATA.artists.multiple,
        message: 'Success',
      });

      await firstStepFn();

      const [artist1, artist2] = TEST_DATA.artists.multiple.map((a) => a.name);
      await actions.selectArtistAndAssertSelected(artist1);
      await actions.selectArtistAndAssertSelected(artist2);

      expect(screen.getByText(artist1)).toBeInTheDocument();
      expect(screen.getByText(artist2)).toBeInTheDocument();

      const removeButton = screen.getByLabelText(
        `steps.step2.removeArtist ${artist1}`
      );
      await user.click(removeButton);
      expect(screen.queryByText(artist1)).not.toBeInTheDocument();
      expect(screen.getByText(artist2)).toBeInTheDocument();
    });
````

DanielMoraDC: The other tests I believe are ok to be in each section as they are specific to the flow.

src/components/generate/usePlaylistSubmission.tsx
barroro: We could use a `SubmissionStatus` to manage translations, and the Stepper component could decide the message based on the status. Unfortunately, we can't implement the mapper that you have because we are not receiving the status in the response. I'm going to make some changes based on this approach.

barroro: Good point! I'm going to try to implement and upload the changes. 👌

barroro: Done! I've already uploaded the changes 👌

src/components/generate/usePlaylistSubmission.tsx
barroro: Done!

__tests__/generate.test.tsx
barroro: If we do it this way, then it doesn't make sense to have the separation by `describe`, does it? I made that separation to make the tests more readable.

DanielMoraDC: I think we can keep a top-level describe for the generate process with the tests that are parametrized. Then a describe for each flow that only contains the tests that are specific for its flow.

src/components/generate/GeneratePlaylistStepper.tsx
barroro: Yes, what you said is true, but we always have to define all the schemas. We just modify them to indicate whether they are optional or not. However, they have to exist in the schema.

barroro: Done! I was wrong, you were absolutely right.

DanielMoraDC (approved):
Great job 🚀

PR #41, opened 2025-05-10
fix: behavior of validation errors

barroro:
## Description
Add several fixes related to the behavior of validations in the form.
- Remove the native `required` attribute from HTML
image

- Hide validation error in artists selector when the value is valid after adding or removing any artist.

image


DanielMoraDC:
I do not know how to review these changes as I am not sure of why these are needed or what is the actual impact? Could you extend the PR description?

DanielMoraDC (approved):

PR #43, opened 2025-05-10
feat: [FW-104] display artists as a list instead of bubbles

barroro:
## Description
Change the display of selected artists by showing each artists image

### DIFF
--- a/src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
+++ b/src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
@@ -8,13 +8,14 @@ interface SelectedArtistBadgeProps {
   onRemove: (_artistName: string) => void;
 }
 
+const IMAGE_SIZE = 24;
+
 const SelectedArtistBadge = ({
   name,
   imageUri,
   onRemove,
 }: SelectedArtistBadgeProps) => {
   const { t } = useTranslation('generate');
-  const IMAGE_SIZE = 24; // Example size, adjust as needed
 
   return (
     <div className="inline-flex items-center overflow-hidden rounded-lg border-2 bg-white">
--- a/src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
+++ b/src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
@@ -1,0 +1,54 @@
+import Image from 'next/image';
+import { X } from 'lucide-react';
+import useTranslation from 'next-translate/useTranslation';
+
+interface SelectedArtistBadgeProps {
+  name: string;
+  imageUri?: string;
+  onRemove: (_artistName: string) => void;
+}
+
+const SelectedArtistBadge = ({
+  name,
+  imageUri,
+  onRemove,
+}: SelectedArtistBadgeProps) => {
+  const { t } = useTranslation('generate');
+  const IMAGE_SIZE = 24; // Example size, adjust as needed
+
+  return (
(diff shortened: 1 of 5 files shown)


src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onRemove: (_artistName: string) => void;
+}
+
+const SelectedArtistBadge = ({
+  name,
+  imageUri,
+  onRemove,
+}: SelectedArtistBadgeProps) => {
+  const { t } = useTranslation('generate');
+  const IMAGE_SIZE = 24; // Example size, adjust as needed
```
DanielMoraDC, line 17: Should we define this outside the scope of the function, instead?

src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { X } from 'lucide-react';
+import useTranslation from 'next-translate/useTranslation';
+
+interface SelectedArtistBadgeProps {
+  name: string;
+  imageUri?: string;
+  onRemove: (_artistName: string) => void;
+}
+
+const SelectedArtistBadge = ({
```
DanielMoraDC, line 11: For consistency I would define this as a named function instead of an arrow function. It is also preferred in the [Google's Typescript Guide](https://google.github.io/styleguide/tsguide.html#function-declarations), which I sometimes use as reference.

DanielMoraDC:
I see they are still horizontally stacked. I think something like the following feels more comfortable:

Captura de pantalla 2025-05-11 a les 12 35 23


What do you think?

DanielMoraDC:
I played around with Copilot and got this:

https://github.com/user-attachments/assets/aaaf115c-429e-4575-9442-63bc8b5a3080

I pushed it into [this branch](https://github.com/festwrap/festwrap-ui/tree/NOJIRA_artist-list-w-scroll-poc). Let me know what you think.

src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
barroro: I agree! I'm going to change it 👌

barroro: Done!

src/components/generate/PlaylistSearchArtistsForm/SelectedArtistBadge.tsx
barroro: What do you mean? We have defined every component with an arrow function 🤔

DanielMoraDC: I see we have many functions defined with arrow functions but then some of them with named functions, too ([example](https://github.com/festwrap/festwrap-ui/blob/main/src/components/ui/Stepper.tsx#L31)). What's the motivation for using arrow functions by default? I see them useful for inline, short functions, but I believe they make code base harder to read (i.e. harder to quickly identify functions from variables).

barroro: In this case, it is incorrect. It should be an arrow function. I don't think I have any particular reason to use it. I implemented it this way because I have been working this way for a long time. The only advantage I see right now is that you have to keep the correct order when you define different functions in the same file. However, I don't think that's a good enough reason to choose this approach.

PR #45, opened 2025-05-15
feat: [FW-103] add form validation to limit the number of artists

barroro:
# Description
Limit the number of artists you can add to create or update a playlist. The limit is for the limit that we have right now with Setlist.FM API.

image

### DIFF
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -305,14 +305,15 @@ describe('GeneratePlaylistPage', () => {
       expect(screen.getByText(artist2)).toBeInTheDocument();
     });
 
-    it('should add a maximum of 5 artists in the selection', async () => {
+    it('should show an error if more than maximum artists selected', async () => {
+      const nArtists = 6;
       const mockedArtists = [
-        ...TEST_DATA.artists.multiple,
-        { name: 'New Artist 1', imageUri: undefined },
-        { name: 'New Artist 2', imageUri: undefined },
-        { name: 'New Artist 3', imageUri: undefined },
-        { name: 'New Artist 4', imageUri: undefined },
-        { name: 'New Artist 5', imageUri: undefined },
+        ...Array(nArtists)
+          .fill(null)
+          .map((_, i) => ({
+            name: 'New Artist ' + i,
+            imageUri: undefined,
+          })),
       ];
 
       vi.mocked(mockServices.artistsService.searchArtists).mockResolvedValue({
@@ -322,17 +323,11 @@ describe('GeneratePlaylistPage', () => {
 
       renderWithProviders(<GeneratePlaylistPage {...staticTranslations} />);
 
-      const [artist1, artist2, artist3, artist4, artist5, artist6] =
-        mockedArtists.map((a) => a.name);
-
       await actions.completeFirstStepNewPlaylist();
 
-      await actions.selectArtistAndAssertSelected(artist1);
-      await actions.selectArtistAndAssertSelected(artist2);
-      await actions.selectArtistAndAssertSelected(artist3);
-      await actions.selectArtistAndAssertSelected(artist4);
-      await actions.selectArtistAndAssertSelected(artist5);
-      await actions.selectArtistAndAssertSelected(artist6);
+      for (const artist of mockedArtists) {
+        await actions.selectArtistAndAssertSelected(artist.name);
+      }
 
       expect(
         await screen.findByText(/steps.errors.selectedArtists.max/i)
--- a/__tests__/generate.test.tsx
+++ b/__tests__/generate.test.tsx
@@ -305,6 +305,44 @@ describe('GeneratePlaylistPage', () => {
       expect(screen.getByText(artist2)).toBeInTheDocument();
     });
 
+    it('should add a maximum of 5 artists in the selection', async () => {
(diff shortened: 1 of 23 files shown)


src/components/generate/GeneratePlaylistStepper.tsx
```diff
@@ -31,6 +31,7 @@ const baseSchema = z.object({
         imageUri: z.string().optional(),
       })
     )
+    .max(5, 'steps.errors.selectedArtists.max')
```
DanielMoraDC, line 34: Can we move this to a constant at the top?

locales/ca/generate.json
```diff
@@ -13,6 +13,7 @@
         "required": "Seleccionar una llista de reproducció és obligatori"
       },
       "selectedArtists": {
+        "max": "Només es poden afegir 5 artistes a la llista de reproducció",
```
DanielMoraDC, line 16: I do not like the fact that we couple the number of artists to the actual message. Can we add it dynamically?

__tests__/generate.test.tsx
```diff
@@ -305,6 +305,44 @@ describe('GeneratePlaylistPage', () => {
       expect(screen.getByText(artist2)).toBeInTheDocument();
     });
 
+    it('should add a maximum of 5 artists in the selection', async () => {
```
DanielMoraDC, line 308: I believe we can simplify and reuse some code in the test. Here is a proposal:

```typescript
it('should show an error if more than maximum artists selected', async () => {
      const nArtists = 6;
      const mockedArtists = [
        ...Array(nArtists).fill(null).map((_, i) => ({ 
          name: "New Artist " + i, 
          imageUri: undefined 
        }))
      ];

      vi.mocked(mockServices.artistsService.searchArtists).mockResolvedValue({
        artists: mockedArtists,
        message: 'Success',
      });

      await actions.completeFirstStepNewPlaylist();

      for (const artist of mockedArtists) {
        await actions.selectArtistAndAssertSelected(artist.name);
      }

      expect(
        await screen.findByText(/steps.errors.selectedArtists.max/i)
      ).toBeInTheDocument();

      expect(
        mockServices.playlistsService.createNewPlaylist
      ).not.toHaveBeenCalled();
    });
```

I basically changed:
- Removed the test artists in `TEST_DATA` as they are not needed in the test.
- Used a map to create the artists to remove duplication.
- Select artists in a loop to remove duplication.

I also changed the name to make it more generic so the artist limit is not hardcoded in the name, too.

DanielMoraDC: Let me know what you think.

barroro: Good approach! I think that is more readable 👌

barroro: Done!

src/components/generate/GeneratePlaylistStepper.tsx
barroro: Sure! I'm going to change it 👌

barroro: Done!

locales/ca/generate.json
barroro: Done!

src/constants/playlist.ts
```diff
@@ -0,0 +1 @@
+export const MAX_ARTISTS = 5;
```
DanielMoraDC, line 1: This breaks consistency with the current use of constants in the app. I would rather move all constants to this file or keep the constant [here](https://github.com/festwrap/festwrap-ui/pull/45/files#diff-2b99fe87a4cad22562978a7d7de0925132038ffe624113ef5642ad9893dd03e1L19) and do this work in a separate PR.

barroro: It's a bit different in this case. If you are comparing with `IMAGE_SIZE` or any constant in the UI layer, they are not the same because this is a validation logic shared between components and it could be part of the domain logic, while the `IMAGE_SIZE` is UI logic related to the images styles instead.
We could move to the file that you commented but I don't like to use constants between components and are located in one of these files, because we can create a circular dependency.

DanielMoraDC: > It's a bit different in this case. If you are comparing with IMAGE_SIZE or any constant in the UI layer, they are not the same because this is a validation logic shared between components and it could be part of the domain logic, while the IMAGE_SIZE is UI logic related to the images styles instead.

In that case, we could simply either create different constant files per layer or use different sections in the file, per layer. Does that make sense?

> We could move to the file that you commented but I don't like to use constants between components and are located in one of these files, because we can create a circular dependency.

If we only define constants with simple, native types, we should never create a circular dependency, right?

DanielMoraDC: The main problem I see with this is that we now need to look at two different places to check for those values, which is confusing.

barroro: What do you mean when you talk about the different sections in the file? We could try to move the constant in the main component and import from this component to the another. I wouldn't create a constant file per layer I think that it creates more complexity.
Also we could move this constant to the `playlist.ts` file where they are all the types. What do you think?

DanielMoraDC: Sounds good!

DanielMoraDC (approved):

PR #46, opened 2025-05-26
refactor: [FW-105] simplify generate form removing steps

barroro:
# Description

Simplify the generate form to allow filling and submitting the form on the same page without navigating between steps.
Now, when the form has been successfully submitted, it redirects to a new page where you can see the preview.

# Changes
- Rename and restructure folders based on the changes.
- Remove steps navigation and all the logic related.
- Remove `Stepper` component.
- Add `Alert` component to display the messages in the success page.
- Remove logic related to the steps in the `generate.test` file

# Testing

- Check the Figma design to ensure every part is applied based on the prototypes.
- Check success scenarios when we could add all the artists, or when we couldn't add all.


__tests__/generate-success.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+      'https://open.spotify.com/embed/playlist/test-playlist-id-2'
+    );
+
+    expect(screen.getByRole('alert')).toBeInTheDocument();
+    expect(
+      screen.getByText('errors.submitPlaylist.missingArtists')
+    ).toBeInTheDocument();
+  });
+
+  it('renders without playlistId (should ideally not happen but test robustness)', () => {
```
DanielMoraDC, line 66: I am not sure what this test is checking. Shouldn't we show an error in that case?

__tests__/generate-success.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    render(<PlaylistGeneratedSuccessfully />);
+
+    expect(document.title).toBe('meta.title');
+
+    const spotifyIframe = screen.queryByTitle('Spotify embedded playlist');
+    expect(spotifyIframe).not.toBeInTheDocument();
+
+    expect(screen.queryByRole('alert')).not.toBeInTheDocument();
+  });
+});
```
DanielMoraDC, line 80: Why should also test that the generation ends up with error, right?

i18n.json
```diff
@@ -7,12 +7,8 @@
   ],
   "pages": {
     "*": [
-      "common"
-    ],
-    "/": [
-      "home"
-    ],
-    "/generate": [
+      "common",
```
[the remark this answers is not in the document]
DanielMoraDC: Resolving, I understand now.

DanielMoraDC:
Good job 👏  I tested a bit locally, the flow looks good! I did not go through all the translations but given we need to perform some QA, we can validate that then.

One thing we could do to simplify the experience is to remove the description input and use an empty description for now. in the code. What do you think?

__tests__/generate-success.test.tsx
barroro: It's there. It's the second one 'renders correctly when playlistId and partialError are provided' 🤔

DanielMoraDC: I understand these are covered in the first view. I did not notice that. Resolving!

PR #47, opened 2025-05-29
feat: [FW-102] add beta info alert in generate flow

barroro:
## Description
We need to gather feedback on our application to identify areas for improvement. So, we have decided to add a link to a feedback form so that users can respond.

## Changes
- Create new component in `generate` folder called `BetaInfoAlert`
- Add `BetaInfoAlert` in the two pages of the generator flow
   - Generate form page
   - Success page


DanielMoraDC:
I feel we maybe have to align the message with the form content. The message looks like this:

```text
This site is in beta. You can help us improve the experience by reporting any issues you may encounter
```

But then the forms asks about general opinion. I feel people will only click there if they encounter any problem, which is not the only thing we want to capture.

What about changing it to something like this:

```text
This site is in beta. Help us improve the experience by providing your feedback
```

barroro:
> I feel we maybe have to align the message with the form content. The message looks like this:
> 
> ```
> This site is in beta. You can help us improve the experience by reporting any issues you may encounter
> ```
> 
> But then the forms asks about general opinion. I feel people will only click there if they encounter any problem, which is not the only thing we want to capture.
> 
> What about changing it to something like this:
> 
> ```
> This site is in beta. Help us improve the experience by providing your feedback
> ```

Done! 🚀

DanielMoraDC (approved):
🚀 🎸

PR #48, opened 2025-05-31
fix: use GCP auth on prod by default

DanielMoraDC:
# Description

Using GCP Auth client when deploying in prod.

Also creating single client instances in a shared file so each handler does not need to create their own. By doing so, we are alsoreducing code duplicating and improving maintainability.


barroro (approved):
Good job! It was very clean 🔥

PR #53, opened 2025-06-14
feat: [FW-139] improve load artists

barroro:
## Description
To improve the UX when users are searching artists, we have to give more feedback when the artists are not loaded yet in the selector.

## Changes
- Add loading status in the selector.
- Reduce the delay time in the debounce input.

### DIFF
--- a/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
+++ b/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
@@ -7,7 +7,6 @@ import { Loader2Icon } from 'lucide-react';
 /* eslint-disable no-unused-vars */
 export enum ArtistSearchStatus {
   Searching = 1,
-  Empty,
   Error,
   NoResults,
   HasResults,
@@ -59,8 +58,6 @@ export function ArtistSearchResultList({
             {t('playlistSearchArtists.artistSearch.searching')}
           </li>
         );
-      case ArtistSearchStatus.Empty:
-        return buildStatusItem(t('playlistSearchArtists.artistSearch.empty'));
       case ArtistSearchStatus.Error:
         return buildAlertStatusItem(t('errors.artistSearch.unexpectedError'));
       case ArtistSearchStatus.NoResults:
--- a/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
+++ b/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
@@ -2,6 +2,7 @@ import { ArtistDTO } from '@/entities/artists';
 import ArtistSearchResult from './ArtistSearchResult';
 import useTranslation from 'next-translate/useTranslation';
 import { cn } from '@/lib/utils';
+import { Loader2Icon } from 'lucide-react';
 
 /* eslint-disable no-unused-vars */
 export enum ArtistSearchStatus {
@@ -50,8 +51,11 @@ export function ArtistSearchResultList({
   function renderSearchItems() {
     switch (status) {
       case ArtistSearchStatus.Searching:
-        return buildStatusItem(
-          t('playlistSearchArtists.artistSearch.searching')
+        return (
+          <li className="px-4 py-2 text-secondary flex items-center justify-center">
+            <Loader2Icon className="h-5 w-5 text-secondary animate-spin mr-2" />
+            {t('playlistSearchArtists.artistSearch.searching')}
+          </li>
         );
       case ArtistSearchStatus.Empty:
         return buildStatusItem(t('playlistSearchArtists.artistSearch.empty'));
--- a/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
+++ b/src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
@@ -11,7 +11,6 @@ export enum ArtistSearchStatus {
   Error,
   NoResults,
   HasResults,
-  Loading,
 }
 
 type ArtistSearchResultListProps = {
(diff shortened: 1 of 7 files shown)


src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/PlaylistSearchArtistsSection.tsx
```diff
@@ -30,7 +30,7 @@ const PlaylistSearchArtistsSection = () => {
 
   const debouncedSearch = useDebouncedCallback((searchTerm: string) => {
     search(searchTerm);
-  });
+  }, 300);
```
DanielMoraDC, line 33: Could we move this to a constant? Changing the default value instead of using a magic number here also works.

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
```diff
@@ -49,8 +50,14 @@ export function ArtistSearchResultList({
...
+        return (
+          <li
+            className="px-4 py-2 text-secondary flex items-center justify-center"
+            role="status"
+          >
+            <Loader2Icon className="h-5 w-5 text-secondary animate-spin mr-2" />
+            {t('playlistSearchArtists.artistSearch.searching')}
+          </li>
         );
       case ArtistSearchStatus.Empty:
```
DanielMoraDC, line 62: Do we actually need this, now, if we have the placeholder? I believe it is redundant, now.

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/SearchArtistsCombobox.tsx
```diff
@@ -101,6 +108,7 @@ export function SearchArtistsCombobox({
 
   const clearSearch = () => {
     setSearch('');
+    onSearch('');
```
DanielMoraDC, line 111: It feels inefficient to call the search function with an empty artist. Is that needed?

DanielMoraDC:
Great job. Left some small comments.

Not related to this PR but I thought we can use this chance to fix that (or handle in a separate PR):

https://github.com/user-attachments/assets/0c4be887-31c2-42d7-8f92-9ce69cfa640b

When searching, the "no results" appear even if it is searching. Do you know how we can fix that?

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/SearchArtistsCombobox.tsx
```diff
@@ -140,16 +148,21 @@ export function SearchArtistsCombobox({
             aria-autocomplete="list"
             aria-controls="combobox-items"
             data-testid="artist-search-input"
+            placeholder={placeholder}
           />
-          {search && (
+          {isSearching ? (
```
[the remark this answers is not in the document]
barroro: It's different. This condition allows to display a spinner inside the input on the right side.

barroro:
> Great job. Left some small comments.
> 
> Not related to this PR but I thought we can use this chance to fix that (or handle in a separate PR):
> 
>  Gravacio.de.pantalla.2025-06-17.a.les.23.10.06.mov 
> When searching, the "no results" appear even if it is searching. Do you know how we can fix that?

Yest, I saw the error yesterday. Let me check, and I'm going to try to fix it 👌

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
barroro: I think it is elegant to block the button to delete the input value while searching. 💎

DanielMoraDC: Sorry, I meant the `ArtistSearchStatus.Empty`. I believe when empty, the placeholder is shown and we never get into this condition.

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/SearchArtistsCombobox.tsx
DanielMoraDC: If eel we are depending on an implementation detail anyway. Why do we need to call the onSearch when empty?

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
barroro: True! I'm going to remove it 👌

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/SearchArtistsCombobox.tsx
barroro: The parent component provides the `options` prop (search results). If you don't tell the parent the search was cleared, it might continue showing old results.

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/ArtistSearchResultList.tsx
barroro: Done!

src/components/generate/GeneratePlaylistForm/PlaylistSearchArtistsForm/SearchArtistsCombobox.tsx
DanielMoraDC: In that case, I think we'd better pass and use the `clearArtists` from `useArtistSearch`, instead. Does that make sense?

DanielMoraDC: Actually, I comment `hasSearched` state, it works without the `onSearch('')` 🤔

barroro: Done! I've removed `hasSearched` state 👌

barroro: Done! I've removed `onSearch` from `onClear` method.

DanielMoraDC (approved):

PR #54, opened 2025-06-15
fix: [FW-141] improve ESLint Configuration with TypeScript Support

barroro:
## Description

This PR refactors our existing ESLint configuration to add comprehensive TypeScript support and improve code quality checks.

### Configuration Improvements

- **Enhanced TypeScript support** by adding @typescript-eslint/eslint-plugin@^7.2.0
- **Upgraded base configuration** to use `next/core-web-vitals` for better Next.js integration
- **Added targeted TypeScript rules** for `.ts` and `.tsx` files with overrides
- **Implemented safety rules** to prevent unsafe type operations

### Benefits

- **Enhanced type safety**: Improved TypeScript validation and interface checking
- **Better development experience**: Catches type mismatches during development
- **Optimized for Next.js**: Updated to use recommended Next.js ESLint configuration
- **Targeted linting**: TypeScript rules strategically applied only to relevant file types

### DIFF
--- a/locales/ca/home.json
+++ b/locales/ca/home.json
@@ -4,14 +4,9 @@
     "subtitle": "Afegeix les cançons dels sets més recents que sonaran als teus propers concerts",
     "title": "Crea <0>llistes de Spotify</0> pels teus concerts i festivals"
   },
-  "howItWorksSection": {
-    "meta": {
-      "description": "Llistes de Spotify pels teus concerts i festivals",
-      "keywords": "Spotify, Setlist.fm, Generador de llistes, Festivals, Música, Concerts",
-      "title": "Inici | Festwrap"
-    },
-    "step1": {
-      "title": "Inicia sessió al teu compte de Spotify"
-    }
+  "meta": {
+    "description": "Llistes de Spotify pels teus concerts i festivals",
+    "keywords": "Spotify, Setlist.fm, Generador de llistes, Festivals, Música, Concerts",
+    "title": "Inici | Festwrap"
   }
 }
--- a/locales/en/home.json
+++ b/locales/en/home.json
@@ -4,14 +4,9 @@
     "subtitle": "Add the most recent set songs that will be played in your upcoming shows",
     "title": "Create <0>Spotify playlists</0> for your concerts and festivals"
   },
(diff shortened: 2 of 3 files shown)


src/pages/api/artists/search.test.ts
```diff
@@ -9,13 +9,20 @@ vi.mock('next-auth/jwt', () => ({
   getToken: vi.fn(),
 }));
 
+type SearchQuery = {
```
DanielMoraDC, line 12: It feels wrong that we are defining the entity in the test code. Is this required?

src/pages/api/playlists/create.test.ts
```diff
@@ -15,14 +15,26 @@ vi.mock('next-auth/jwt', () => ({
   getToken: vi.fn(),
 }));
 
+type CreatePlaylistRequestBody = {
```
DanielMoraDC, line 18: Same as in https://github.com/festwrap/festwrap-ui/pull/54/files#r2148581507

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,14 +1,22 @@
 import NextAuth from 'next-auth';
 import SpotifyProvider from 'next-auth/providers/spotify';
+import { JWT } from 'next-auth/jwt';
+import { Account, Session } from 'next-auth';
+
+interface SpotifyJWT extends JWT {
```
Comment (DanielMoraDC) [inline src/pages/api/auth/[...nextauth].ts:6]: Most of our data structures use `type` instead of interface. For consistency, maybe we could do the following, instead:

```typescript
type SpotifyJWT = JWT & {
  accessToken: string;
  refreshToken: string;
  expiresAt: number;
};
```

src/pages/api/artists/search.test.ts
DanielMoraDC: Can we disable that restriction overall? It feels a bit too restrictive. The alternative would be to place this in the `search.ts` file and import from there.

src/pages/api/playlists/create.test.ts
barroro: Done!

DanielMoraDC (approved):
🚀

PR #56, opened 2025-06-17
chore: remove deprecated "how to" section

DanielMoraDC:
# Description

Those steps are no longer aligned with the expected process. Given the process is now much simpler, there is no need to have that section anymore.

I am also restructuing the Footer to make it simpler. Here is the result:

Captura de pantalla 2025-06-17 a les 8 16 41


src/components/layout/Footer/Footer.tsx
```diff
@@ -8,19 +8,13 @@ const Footer = () => {
 
   return (
     <footer className="bg-secondary bg-opacity-30 text-dark">
-      <div className="max-w-7xl mx-auto px-4 sm:px-6 lg:px-8">
```
barroro, line 11: I would keep the same nav header options here for consistency. I don't think it would be a problem to have it

barroro (commented):
Good job! Let me know what you think about keeping the same design footer 👌

src/components/layout/Footer/Footer.tsx
barroro: I'm still seeing the footer different than before. I would refer to keep as before with the same structure etc... 🤔 
image

barroro: Yes, I agree that it looks too big. Okay, let's leave it at that, but we have to fix it for mobile screens. It doesn't look good on mobile screens now.

image

DanielMoraDC: Good catch! These links are empty and they are going to be replaced by a FAQ section, so we can remove them:

Captura de pantalla 2025-06-20 a les 8 31 58

Done ✅

barroro (approved):
Good job!

PR #64, opened 2025-07-28
fix: [FW-153] implement handling error and retries strategy to ensure load iframe correctly

barroro:
## Description
Implement several strategies to ensure that the user can display the playlist created.

### Scenarios
- Loading iframe
image

- Fallback iframe
image

- Retries information before fallback
image

## Changes
- Add new child components for the overview component
- Create a new folder to encapsulate all the logic
- Add test cases


src/features/generate/components/playlist-update-report/use-playlist-iframe-loader.ts
```diff
@@ -0,0 +1,85 @@
...
+  playlistId: string;
+  autoRetryDelayMs?: number;
+  maxRetries?: number;
+}
+
+interface UsePlaylistIframeLoaderReturn {
+  isLoading: boolean;
+  hasError: boolean;
+  retryCount: number;
+  key: number;
```
DanielMoraDC, line 13: nit: I would make some of these variables more specific, as it was not clear what they were at first. For example:
- embedUrl: playlistEmbedUrl.
- directUrl: playlistUrl.
- key: iframeKey

src/features/generate/components/playlist-update-report/use-playlist-iframe-loader.ts
```diff
@@ -0,0 +1,85 @@
...
+  handleIframeError: () => void;
+  handleRetry: () => void;
+  shouldShowFallback: boolean;
+  shouldShowRetryInfo: boolean;
+  shouldShowSuccessButton: boolean;
+}
+
+export const usePlaylistIframeLoader = ({
+  playlistId,
+  autoRetryDelayMs = 3000,
```
DanielMoraDC, line 26: nit: for consistency, maybe we should also move these into constants above.

locales/ca/generate.json
```diff
@@ -77,7 +77,13 @@
   },
   "playlistSuccess": {
     "description": "Visualitza les cançons afegides a a llista",
+    "embedLoadError": "No s'ha pogut carregar la llista",
+    "embedLoadErrorDescription": "La llista de reproducció no s'ha pogut carregar. Pots intentar-ho de nou o obrir-la directament a Spotify.",
+    "loadingPlaylist": "Carregant llista de reproducció...",
+    "openInSpotify": "Obrir a Spotify",
     "playlistGeneratedSuccessfully": "Llista generada amb èxit",
+    "retryEmbed": "Provar de nou",
+    "retryingLoad": "Reintentant carregar",
```
DanielMoraDC, line 86: nit: this sounds a bit unnatural. Maybe "Provant de nou..."?

